PR #2, opened 2016-06-01
Adding AttributeMeta, StringMeta and MapMeta

GDYendell:
How is this?

### DIFF
--- a/malcolm/core/mapmeta.py
+++ b/malcolm/core/mapmeta.py
@@ -26,8 +26,8 @@ class MapMeta(Loggable):
             ValueError:
         """
 
-        if attribute_meta.name in self.elements.keys:
-            raise ValueError("Element already exists in map")
+        if attribute_meta.name in self.elements.keys():
+            raise ValueError("Element already exists in dictionary")
         else:
             self.elements[attribute_meta.name] = attribute_meta
             self.required.append(required)
--- a/malcolm/core/mapmeta.py
+++ b/malcolm/core/mapmeta.py
@@ -9,7 +9,7 @@ class MapMeta(Loggable):
         super(MapMeta, self).__init__(logger_name=name)
 
         self.name = name
-        self.elements = []
+        self.elements = {}
 
     def add_element(self, attribute_meta, required=False):
         """
@@ -20,4 +20,4 @@ class MapMeta(Loggable):
             required(bool): Whether attribute is required or optional
         """
 
-        self.elements.append((attribute_meta, required))
+        self.elements[attribute_meta.name] = (attribute_meta, required)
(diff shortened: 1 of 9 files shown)


GDYendell:
Is the StringMeta validate() too restrictive?

malcolm/core/mapmeta.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class MapMeta(Loggable):
+    """A meta object to store a set of attribute metas"""
+
+    def __init__(self, name):
+        super(MapMeta, self).__init__(logger_name=name)
+
+        self.name = name
+        self.elements = OrderedDict({})
```
thomascobb: You don't need to pass {} to OrderedDict

malcolm/core/mapmeta.py
```diff
@@ -0,0 +1,25 @@
...
+    def add_element(self, attribute_meta, required=False):
+        """
+        Add an element, stating whether it is required.
+
+        Args:
+            attribute_meta(AttributeMeta): Attribute instance to store
+            required(bool): Whether attribute is required or optional
+        """
+
+        self.elements[attribute_meta.name] = (attribute_meta, required)
```
thomascobb: Can you make the self.required a list of strings so it's easy to serialise?

thomascobb:
I was a little more permissive here:
https://github.com/dls-controls/malcolm/blob/master/malcolm/core/vtype.py#L179

Just less lines of code, don't really mind either way

tests/test_core/test_attributemeta.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def setUp(self):
+        self.attribute_meta = AttributeMeta("Test")
+
+    def test_given_validate_called_then_raise_error(self):
+
+        expected_error_message = \
+            "Abstract validate function must be implemented in child classes"
+
+        with self.assertRaises(NotImplementedError) as error:
```
thomascobb: Didn't know you could do this!

tests/test_core/test_stringmeta.py
```diff
@@ -0,0 +1,48 @@
...
+
+        self.assertEqual("15", response)
+
+    def test_given_value_float_then_cast_and_set(self):
+        response = self.string_meta.validate(12.8)
+
+        self.assertEqual("12.8", response)
+
+    def test_given_value_dict_then_raise_error(self):
+        expected_error_message = "Value must be of type str or castable to str"
```
thomascobb: Although checking the error message is probably a bit too much work...

malcolm/core/mapmeta.py
```diff
@@ -5,21 +5,23 @@
...
-        Add an element, stating whether it is required.
+        Add an element and whether it is required.
 
         Args:
             attribute_meta(AttributeMeta): Attribute instance to store
             required(bool): Whether attribute is required or optional
         """
 
-        self.elements[attribute_meta.name] = (attribute_meta, required)
+        self.elements[attribute_meta.name] = attribute_meta
```
thomascobb: What happens if we add an attribute_meta which is already in the map?

PR #4, opened 2016-06-01
Add initial Method implementation

xen-0:


malcolm/core/method.py
```diff
@@ -0,0 +1,66 @@
...
+
+from loggable import Loggable
+
+class Method(Loggable):
+    """Exposes a function with metadata for arguments and return values"""
+
+    def __init__(self, name):
+        self.name = name
+        self._func = None
+        self._args = OrderedDict()
```
thomascobb: Can we make this self.takes so serializing will be easier

malcolm/core/method.py
```diff
@@ -0,0 +1,66 @@
...
+        self._returns = None
```
thomascobb: And self.returns

malcolm/core/method.py
```diff
@@ -0,0 +1,54 @@
...
+    def set_function_returns(self, return_meta):
+        """Set the return parameters for the method to validate against"""
+        self.returns = return_meta
+
+    def __call__(self, *args, **kwargs):
+        """Call the exposed function using regular keyword argument parameters.
+        Will validate the output against provided return parameters.
+        """
+
+        for arg in self.takes.elements:
```
thomascobb: What about the positional args?

malcolm/core/method.py
```diff
@@ -0,0 +1,54 @@
...
+            if arg not in kwargs.keys():
+                if arg in self.defaults.keys():
+                    kwargs[arg] = self.defaults[arg]
+                elif arg in self.takes.required:
+                    raise ValueError("Argument %s is required but was not provided" % arg)
+        return_val = self.func(kwargs)
```
thomascobb: Mine was a bit longer...

https://github.com/dls-controls/malcolm/blob/master/malcolm/core/method.py#L151

PR #12, opened 2016-06-02
Add Response class

xen-0:


malcolm/core/response.py
```diff
@@ -0,0 +1,32 @@
...
+    @classmethod
+    def Return(cls, id_, context, value=None):
+        """Create a Return response object with the provided parameters.
+
+        Args:
+            id (int): id from initial message
+            context: context associated with id
+            value: object return value (default None)
+        """
+        response = Response(id_, context, "Return")
```
thomascobb: Can you use cls here instead of Response

PR #16, opened 2016-06-02
Create Request class + add Block handle_request function

GDYendell:
Also simplified Block to_dict test.


malcolm/core/request.py
```diff
@@ -0,0 +1,74 @@
+from collections import OrderedDict
+from malcolm.core.response import Response
+
+
+class Request(object):
+    """An object to interact with the attributes of a Block"""
+
+    POST = "Post"
```
thomascobb: I think we decided we don't need this...

tests/test_core/test_request.py
```diff
@@ -0,0 +1,74 @@
...
+        param_dict = dict(one=7, two=23)
+        post = Request.Post(3, self.context, self.response_queue, [""], parameters=param_dict)
+
+        self.assertEqual(param_dict, post.parameters)
+
+    def test_given_invalid_attr_then_raise_error(self):
+        param_dict = dict(one=7, two=23)
+        post = Request.Post(3, self.context, self.response_queue, [""], parameters=param_dict)
+
+        with self.assertRaises(KeyError):
```
thomascobb: Cool, I never knew you could do this...

PR #18, opened 2016-06-03
Added system test for hello controller with process

thomascobb:


tests/test_core/test_system_core.py
```diff
@@ -3,12 +3,19 @@
 import os
 sys.path.append(os.path.join(os.path.dirname(__file__), "..", ".."))
 
+# logging
```
xen-0: Why the commented imports?

thomascobb: Because it's amazing how often I end up having to add debugging in unit tests to work out what is going wrong...

tests/test_core/test_system_core.py
```diff
@@ -37,13 +44,30 @@ def create_methods(self):
...
+        block = Block("hello")
+        HelloController(block)
+        process.add_block(block)
+        process.start()
+        q = sync_factory.create_queue()
+        req = Request.Post(response_queue=q, context="ClientConnection", id_=44,
+                           endpoint=["hello", "say_hello"],
+                           parameters=dict(name="thing"))
+        process.q.put(req)
+        resp = q.get(timeout=1)
```
GDYendell: What is the timeout for?

thomascobb: So that the test fails rather than hangs if something went wrong

PR #19, opened 2016-06-03
Add Get Request and Get handling in block

GDYendell:


malcolm/core/block.py
```diff
@@ -27,8 +27,14 @@ def add_method(self, method):
         setattr(self, method.name, method)
 
     def handle_request(self, request):
-        method_name = request.endpoint[-1]
-        response = self._methods[method_name].handle_request(request)
+        if request.type == "Post":
```
thomascobb: Ok, now we need request.POST = "Post" and request.GET = "Get"

malcolm/core/block.py
```diff
@@ -27,8 +27,14 @@ def add_method(self, method):
...
+            method_name = request.endpoint[-1]
+            response = self._methods[method_name].handle_request(request)
+        else:
+            layer = self
+            for next_link in request.endpoint[1:]:
+                layer = getattr(layer, next_link)
+            response = layer
```
thomascobb: We should call to_dict() on the layer here...

malcolm/core/block.py
```diff
@@ -27,8 +27,14 @@ def add_method(self, method):
...
 
         return response
```
thomascobb: Handle response shouldn't return anything

malcolm/core/block.py
thomascobb: This should be:
`request.respond_with_return(layer.to_dict())`

malcolm/core/block.py
```diff
@@ -27,8 +27,14 @@ def add_method(self, method):
         setattr(self, method.name, method)
 
     def handle_request(self, request):
-        method_name = request.endpoint[-1]
-        response = self._methods[method_name].handle_request(request)
+        if request.type == "Post":
+            method_name = request.endpoint[-1]
+            response = self._methods[method_name].handle_request(request)
```
thomascobb: Handle response shouldn't return anything

malcolm/core/block.py
```diff
@@ -27,18 +27,29 @@ def add_method(self, method):
...
+
+        Args:
+            request(Request): Request object specifying action
+        """
+
+        if request.type == request.POST:
             method_name = request.endpoint[-1]
             response = self._methods[method_name].handle_request(request)
+            request.respond_with_return(response.to_dict())
         else:
```
thomascobb: elif request.type == request.GET:
...
else:
raise

PR #20, opened 2016-06-03
Process spawn

thomascobb:


tests/test_core/test_system_core.py
```diff
@@ -3,12 +3,19 @@
 import os
 sys.path.append(os.path.join(os.path.dirname(__file__), "..", ".."))
 
+# logging
+# import logging
+# logging.basicConfig(level=logging.DEBUG)
+
```
xen-0: Why the commented out imports?

PR #32, opened 2016-06-06
Added ClientController

thomascobb:


malcolm/controllers/clientcontroller.py
```diff
@@ -0,0 +1,61 @@
...
+    def create_methods(self):
+        """Get methods from remote block and mirror them internally"""
+        request = Request.Get(None, None, self.q, [self.block.name])
+        self.client_comms.q.put(request)
+        self.log_debug("Waiting for response to Get %s", self.block.name)
+        response = self.q.get()
+        # Find all the methods
+        for aname, amap in response.value.items():
+            # TODO: If it has "takes" it's a method, flaky...
+            if "takes" in amap:
```
GDYendell: Could you check that 'takes' and/or 'returns' is a MapMeta instance?

thomascobb: It's still a dictionary at this point, so not yet...

PR #36, opened 2016-06-06
Create Error Response class and add functionality for its use...

GDYendell:
...in Request and Method

Also remove lint.


malcolm/core/request.py
```diff
@@ -36,7 +36,7 @@ def set_id(self, id_):
 
     def respond_with_return(self, value=None):
         """
-        Create a Response object to handle the request
+        Create a Return object to handle the request
```
thomascobb: I think Response was correct...

PR #40, opened 2016-06-07
Add Method decorators for improved Controller API

GDYendell:


malcolm/core/method.py
```diff
@@ -106,3 +106,74 @@ def from_dict(cls, name, d):
...
+    Returns:
+        function: Updated function
+    """
+
+    def decorator(func):
+
+        if not hasattr(func, "Method"):
+            Method.wrap_method(func)
+
+        func.Method.takes = args
```
thomascobb: should create map_meta = MapMeta("takes"), then for each pair of args a1, a2 call map_meta.add_element(a1, a2 is not OPTIONAL). If a2 is not in (REQUIRED, OPTIONAL), defaults[a1.name] = a2. Then call func.Method.set_function_takes(map_meta, defaults)

malcolm/core/method.py
```diff
@@ -106,3 +106,74 @@ def from_dict(cls, name, d):
...
+        function: Updated function
+    """
+
+    def decorator(func):
+
+        if not hasattr(func, "Method"):
+            Method.wrap_method(func)
+
+        func.Method.returns = args
+
```
thomascobb: Similar for returns but if a2 is not in (REQUIRED, OPTIONAL) then it is an error

GDYendell:
The patch on TestMethod.test_decorators doesn't work for some reason, so it checks for a MapMeta instance. I want it to check that it is the mock I tried to set up but it seems to be ignoring them. Any idea why this is happening?

GDYendell:
Coveralls implies the test isn't calling the decorator functions properly, which I guess is the problem. Not sure why though...

thomascobb:
Can you rebase onto master please?

GDYendell:
I need to remember to rebase to origin/master...

I think I know how to test it now so I will fix the coverage now.

GDYendell:
OK, how is that?

PR #43, opened 2016-06-08
Create Monitorable class

xen-0:


malcolm/core/monitorable.py
```diff
@@ -0,0 +1,25 @@
...
+        """Sets the parent for changes to be propagated to"""
+        self._logger_name = "%s.%s" % (parent.name, self.name)
+        self.parent = parent
+
+    def on_changed(self, changes):
+        """Propagate changes to parent, adding self.name to paths.
+
+        Args:
+            changes: list of [[path], value] pairs for changed values
+        """
```
thomascobb: We should only do this if we have self.parent != None

PR #44, opened 2016-06-08
Add Attribute class

xen-0:


malcolm/core/attribute.py
```diff
@@ -0,0 +1,45 @@
...
+        self.put_func(value)
+
+    def set_value(self, value):
+        self.value = value
+
+    def to_dict(self):
+        """Create ordered dictionary representing class instance"""
+        d = OrderedDict()
+        d["value"] = self.value
+        d["meta"] = self.meta.to_dict() if self.meta is not None else None
```
thomascobb: meta should never be None

malcolm/core/attribute.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def set_put_function(self, func):
+        self.put_func = func
+
+    def put(self, value):
+        """Call the put function with the given value"""
+        self.put_func(value)
+
+    def set_value(self, value):
+        self.value = value
```
thomascobb: Should call self.on_changed([["value"], value])

PR #46, opened 2016-06-09
Subscribing

xen-0:


malcolm/core/process.py
```diff
@@ -28,12 +36,20 @@ def recv_loop(self):
             if request is PROCESS_STOP:
                 # Got the sentinel, stop immediately
                 break
-            try:
-                self.handle_request(request)
-            except Exception:
-                # TODO: request.respond_with_error()
-                self.log_exception("Exception while handling %s",
-                                   request.to_dict())
+            elif isinstance(request, BlockNotify):
```
thomascobb: How about using a function dictionary as a switch statement:
```python
self.handle_functions = {
    Request.POST: self._forward_block_request,
    Request.PUT: self._forward_block_request,
    Request.SUBSCRIBE: self._handle_subscribe,
    BlockNotify.type_: self._handle_block_notify
}
```
Of course you have to do BlockNotify.type_ = "BlockNotify" somewhere...
Then you can use it like:
```python
try:
    self.handle_functions[request.type_](request)
except Exception:
    self.log_exception(...)
```

malcolm/core/process.py
```diff
@@ -91,3 +108,52 @@ def spawn(self, function, *args, **kwargs):
...
+        for path, value in request.changes:
+            # update changes
+            for e in self._last_changes:
+                if e[0] == path:
+                    e[1] = value
+                    break
+            else:
+                self._last_changes.append([path, value])
+
+            # update cached dict
```
thomascobb: I think this code belongs in block_notify, so that anyone doing a get gets the last consistent serialized structure

malcolm/core/process.py
```diff
@@ -28,12 +36,20 @@ def recv_loop(self):
...
-                # TODO: request.respond_with_error()
-                self.log_exception("Exception while handling %s",
-                                   request.to_dict())
+            elif isinstance(request, BlockNotify):
+                self._handle_block_notify(request)
+            elif isinstance(request, BlockChanged):
+                self._handle_block_changed(request)
+            elif isinstance(request, Request) and \
+                    request.type_ == Request.SUBSCRIBE:
+                self._subscriptions.append(request)
```
thomascobb: Subscribe, and send the serialized current value

malcolm/core/process.py
```diff
@@ -28,12 +36,20 @@ def recv_loop(self):
...
+            else:
+                try:
```
thomascobb: Need to support Get here instead of in Block

malcolm/core/process.py
```diff
@@ -91,3 +108,52 @@ def spawn(self, function, *args, **kwargs):
...
+        for subscription in self._subscriptions:
+            # find stuff that's changed that is relevant to this subscriber
+            endpoint = subscription.endpoint
+            changes = []
+            if len(endpoint) == 0:
+                changes = self._last_changes
+            else:
+                for change in self._last_changes:
+                    i = 0
+                    for (change_path, sub_path) in zip(change[0], endpoint):
```
thomascobb: ok, I give up, what does this do?

xen-0:
The client controller needs Get to be handled (it isn't), and WSServerComms calls process.handle_request explicitly (which doesn't exist) - shouldn't it be putting the request on the process' queue?

malcolm/core/process.py
```diff
@@ -91,3 +112,82 @@ def spawn(self, function, *args, **kwargs):
...
+                    update = update[p]
+                response = Response.Update(
+                    subscription.id_, subscription.context, update)
+                subscription.response_queue.put(response)
+
+    def _handle_block_changed(self, request):
+        """Record changes to made to a block"""
+        for path, value in request.changes:
+            # update changes
+            for e in self._last_changes:
```
thomascobb: How about storing this as:
```python
block_changes = self._last_changes.setdefault(path[0], [])
for e in block_changes:
    if e[0] == path:
        e[1] = value
        break
else:
        block_changes.append([path, value])
```

malcolm/core/process.py
```diff
@@ -91,3 +109,84 @@ def spawn(self, function, *args, **kwargs):
...
+                    for p in endpoint:
+                        update = update[p]
+                    response = Response.Update(
+                        subscription.id_, subscription.context, update)
+                    subscription.response_queue.put(response)
+        self._last_changes[request.name] = []
+
+    def _handle_block_changed(self, request):
+        """Record changes to made to a block"""
+        for path, value in request.changes:
```
thomascobb: Consider:
[["a"],{"b": 32}], [["a", "b"], 42], [["a"], {"b": 52}]

In your code this is compressed to: 
[["a"],{"b": 52}], [["a", "b"], 42]

Which gives a different result.

I think we shouldn't bother with squashing duplicate updates, just pass them through to the client...

malcolm/core/process.py
```diff
@@ -91,3 +109,84 @@ def spawn(self, function, *args, **kwargs):
...
+                for (cp_element, ep_element) in zip(change_path, endpoint):
+                    if cp_element != ep_element:
+                        break
+                    i += 1
+                else:
+                    # change has matching path, so keep it
+                    # but strip off the end point path
+                    filtered_change = [change_path[i:], change_value]
+                    changes.append(filtered_change)
+            if len(changes) > 0:
```
thomascobb: Neat!

malcolm/core/process.py
```diff
@@ -91,3 +109,84 @@ def spawn(self, function, *args, **kwargs):
...
+
+    def _handle_subscribe(self, request):
+        """Add a new subscriber and respond with the current
+        sub-structure state"""
+        subs = self._subscriptions.setdefault(request.endpoint[0], [])
+        subs.append(request)
+        d = self._block_state_cache
+        for p in request.endpoint:
+            d = d[p]
+        response = Response.Update(request.id_, request.context, d)
```
thomascobb: Need a Response.Delta instead of a Response.Update if we asked for it...

tests/test_wscomms/test_system_wscomms.py
```diff
@@ -35,6 +35,7 @@ def setUp(self):
         block = Block("hello")
         self.process.add_block(block)
         HelloController(block)
+        self.process.notify_subscribers(block.name)
```
thomascobb: Why did we need to add this line?

PR #47, opened 2016-06-10
Add support for Python 3.5.

willrogers:
This is a bit tongue-in-cheek.

But don't you want to be supporting python 3?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ sudo: false
 language: python
 python:
   - "2.7"
+  - "3.5"
 
 cache:
   directories:


thomascobb:
I quite like the idea of supporting python 3.5, would you suggest using six or something like that, or would you roll your own compat thing?

willrogers:
I used six, but I haven't done this on a large project.

You don't seem to have many requirements; they should all support python 3 (even cothread, but you can't use that on Travis, I find).

GDYendell:
You may need to update landscape for python 3 checks:

https://docs.landscape.io/configuration.html#python-targets-2-3

thomascobb:
Changes integrated in #51

PR #49, opened 2016-06-14
Change Post responses to go through the process queue first.

xen-0:
Fixes out-of-order messages, where an Update (triggered by a method call
that changes attributes) can arrive after the method's Return message.


malcolm/core/method.py
```diff
@@ -79,10 +80,10 @@ def handle_request(self, request):
         except Exception as error:
             self.log_debug("Error raised %s", error.message)
             message = "Method %s raised an error: %s" % (self.name, error.message)
-            request.respond_with_error(message)
```
thomascobb: I guess you can delete Request.respond_with_* now...

PR #50, opened 2016-06-16
Meta changes

xen-0:
Refactor meta subclass registration and add initial NumberMeta implementation

### DIFF
--- a/malcolm/core/attributemeta.py
+++ b/malcolm/core/attributemeta.py
@@ -7,7 +7,7 @@ class AttributeMeta(Monitorable):
 
     # This will be set by subclasses calling cls.register_subclass()
     metaOf = None
-    # dict mapping metaOf name -> cls
+    # dict mapping metaOf name -> (cls, args)
     _subcls_lookup = {}
 
     def __init__(self, name, description):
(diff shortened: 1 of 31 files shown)


malcolm/core/attributemeta.py
```diff
@@ -35,15 +35,17 @@ def to_dict(self):
         return d
 
     @classmethod
-    def register_subclass(cls, subcls, metaOf):
+    def register_subclass(cls, metaOf, *args):
         """Register a subclass so from_dict() works
 
         Args:
             subcls (AttributeMeta): AttributeMeta subclass to register
```
thomascobb: docs need updating too...

malcolm/core/numbermeta.py
```diff
@@ -0,0 +1,40 @@
+from malcolm.core.attributemeta import AttributeMeta
+
+from pkg_resources import require
+require("numpy")
+import numpy as np
+
+
+_metas = {
```
thomascobb: (matter of taste) I probably wouldn't use an intermediate dictionary, I think it's probably easier to read if you just put the strings straight into the decorator calls...

malcolm/core/numbermeta.py
```diff
@@ -0,0 +1,40 @@
...
+
+@AttributeMeta.register_subclass(_metas[np.int32], np.int32)
+@AttributeMeta.register_subclass(_metas[np.float64], np.float64)
+class NumberMeta(AttributeMeta):
+    """Meta object containing information for a numerical value"""
+
+    def __init__(self, name, description, dtype):
+        super(NumberMeta, self).__init__(name=name, description=description)
+        self.dtype = dtype
+        # TODO where and when should this be set?
```
thomascobb: I think the base class should do this. If we maintain the inverse mapping from (subcls, args) -> metaOf is AttributeMeta.register_subclass, then AttributeMeta.__init__(name, description, *args) can set self.metaOf based on this information

malcolm/core/numbermeta.py
```diff
@@ -0,0 +1,42 @@
+from malcolm.core.attributemeta import AttributeMeta
+from malcolm.compat import base_string
+
+from pkg_resources import require
```
thomascobb: No requires in the source code, just in the test entry points please...

xen-0:
Ok, so this pull-request is starting to include a lot of tangential changes...

PR #59, opened 2016-06-21
Countercontroller system test

xen-0:


malcolm/controllers/countercontroller.py
```diff
@@ -14,9 +14,13 @@ def create_attributes(self):
         yield self.counter
 
     @takes()
-    def reset(self):
+    @returns()
+    def reset(self, args=None):
         self.counter.set_value(0)
+        return {}
```
thomascobb: Can we get away with not having to return an empty dict?

malcolm/controllers/countercontroller.py
```diff
@@ -14,9 +14,13 @@ def create_attributes(self):
         yield self.counter
 
     @takes()
-    def reset(self):
+    @returns()
```
thomascobb: Shouldn't have to do `@returns()` for no return value, maybe Method needs a default empty return map?

PR #61, opened 2016-06-21
Fix block to handle Put requests properly

xen-0:
Similar to method post requests, push a "to return" message to the
process queue so the return is received *after* update to the attribute
has been seen.

### DIFF
--- a/malcolm/core/block.py
+++ b/malcolm/core/block.py
@@ -3,6 +3,8 @@ from contextlib import contextmanager
 
 from malcolm.core.monitorable import Monitorable
 from malcolm.core.process import BlockRespond
+from malcolm.core.request import Request
+from malcolm.core.response import Response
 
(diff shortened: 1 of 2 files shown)


malcolm/core/block.py
```diff
@@ -54,20 +56,18 @@ def handle_request(self, request):
...
-                layer = getattr(layer, next_link)
-
-            if hasattr(layer, "to_dict"):
-                request.respond_with_return(layer.to_dict())
-            else:
-                request.respond_with_return(layer)
+        elif request.type_ == Request.PUT:
+            attr_name = request.endpoint[-1]
+            self._attributes[attr_name].put(request.value)
+            self._attributes[attr_name].set_value(request.value)
```
thomascobb: Don't want to set_value here, the put function will do it if needed

PR #65, opened 2016-06-23
Made Process.block_add happen in process thread

thomascobb:


xen-0:
Do we want to add a test to verify the dispatch to _handle_block_add? That is add_block with the loop running really will add the block.

PR #68, opened 2016-06-27
AttributeMeta developments

xen-0:


malcolm/core/stringarraymeta.py
```diff
@@ -0,0 +1,43 @@
...
+            value (iterable): value to be validated
+
+        Returns:
+            List of Strings or None if value is None
+        """
+        if value is None:
+            return None
+        assert hasattr(value, "__iter__"), \
+            "%s is not iterable" % value
+        if None in value:
```
thomascobb: If you've passed an iterable, then haven't you just exhausted it here?

PR #69, opened 2016-06-27
Add block_response helper method to process

xen-0:
Avoids Block needing to know about Process' internal BlockRespond
message.


malcolm/core/process.py
```diff
@@ -203,6 +203,9 @@ def _handle_get(self, request):
         response = Response.Return(request.id_, request.context, d)
         request.response_queue.put(response)
 
+    def make_block_response_message(self, response, response_queue):
+        return BlockRespond(response, response_queue)
```
thomascobb: Shouldn't it put the message on the q too so it looks like the method below?

PR #71, opened 2016-06-28
Serializable

xen-0:


malcolm/core/serializable.py
```diff
@@ -0,0 +1,72 @@
...
+        typeid = d["typeid"]
+        subcls, args = cls._subcls_lookup[typeid]
+        assert subcls is not cls, \
+            "Subclass %s did not redefine from_dict" % subcls
+        deserialized = subcls.from_dict(name, d, *args)
+        deserialized.typeid = typeid
+        return deserialized
+
+    @classmethod
+    def register_subclass(cls, typeid, *args):
```
thomascobb: Was going to suggest we just call this register()

PR #72, opened 2016-06-29
Block.update

xen-0:


malcolm/core/block.py
```diff
@@ -95,6 +124,8 @@ def to_dict(self):
         for method_name, method in self._methods.items():
             d[method_name] = method.to_dict()
 
+        d["typeid"] = "malcolm:core/Block:1.0"
```
thomascobb: d["typeid"] = self.typeid (isn't it set by Serializable.__init__?)

PR #80, opened 2016-07-11
More metas

GDYendell:
Needs spg-fix first

### DIFF
--- a/malcolm/core/choicearraymeta.py
+++ b/malcolm/core/choicearraymeta.py
@@ -1,0 +1,70 @@
+from collections import OrderedDict
+
+from malcolm.core.scalarmeta import ScalarMeta
+from malcolm.core.serializable import Serializable
+
+
+@Serializable.register("malcolm:core/ChoiceArrayMeta:1.0", "choices_list")
+class ChoiceArrayMeta(ScalarMeta):
+    """Meta object containing information for a choice array"""
+
+    def __init__(self, name, description, choices_list):
+        super(ChoiceArrayMeta, self).__init__(
+            name, description, "choices_list")
+
+        self.choices_list = choices_list  # A list of lists of choices
+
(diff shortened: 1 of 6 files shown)


malcolm/core/choicearraymeta.py
```diff
@@ -0,0 +1,70 @@
...
+from malcolm.core.serializable import Serializable
+
+
+@Serializable.register("malcolm:core/ChoiceArrayMeta:1.0", "choices_list")
+class ChoiceArrayMeta(ScalarMeta):
+    """Meta object containing information for a choice array"""
+
+    def __init__(self, name, description, choices_list):
+        super(ChoiceArrayMeta, self).__init__(
+            name, description, "choices_list")
```
thomascobb: What does "choices_list" passed up do?

malcolm/core/numberarraymeta.py
```diff
@@ -0,0 +1,49 @@
...
+from malcolm.compat import base_string
+
+
+@Serializable.register("malcolm:core/NumberArrayMeta:1.0")
+class NumberArrayMeta(ScalarMeta):
+    """Meta object containing information for an array of numerical values"""
+
+    def __init__(self, name, description, dtype_list):
+        super(NumberArrayMeta, self).__init__(name, description)
+        self.dtype_list = dtype_list
```
thomascobb: Same thing, every element should be the same dtype

malcolm/core/numberarraymeta.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def validate(self, value):
+        if value is None:
```
thomascobb: The output should be a numpy array.

Something like this:
https://github.com/dls-controls/malcolm/blob/master/malcolm/core/vtype.py#L84

PR #82, opened 2016-07-11
Make all controllers take process

thomascobb:


malcolm/controllers/clientcontroller.py
```diff
@@ -75,8 +74,7 @@ def call_server_method(self, method_name, parameters, returns):
         request = Request.Post(None, q,
                                [self.block.name, method_name], parameters)
         self.client_comms.q.put(request)
-        with self.block.lock_released():
-            response = q.get()
+        response = q.get()
```
GDYendell: Don't we still need the lock here?

PR #88, opened 2016-07-13
Create Part and add to Controller

xen-0:


malcolm/core/part.py
```diff
@@ -0,0 +1,10 @@
+from malcolm.core.loggable import Loggable
+
+class Part(Loggable):
+
+    def __init__(self, name, process, block):
+        self.set_logger_name("Part")
```
thomascobb: Think this line is bogus...

PR #90, opened 2016-07-13
Create Hook class

GDYendell:
(+ Fix doc string)


malcolm/core/hook.py
```diff
@@ -0,0 +1,62 @@
...
+class Task(object):
+
+    def __init__(self, process):
+        self.process = process
+
+
+class Hook(Loggable):
+
+    def __call__(self, func):
+        func.hook = self
```
thomascobb: can we make it Hook to be consistent with Method?

PR #92, opened 2016-07-14
Refactor Request and Response to be Serializable

GDYendell:


malcolm/core/process.py
```diff
@@ -62,7 +63,7 @@ def recv_loop(self):
                 # Got the sentinel, stop immediately
                 break
             try:
-                self._handle_functions[request.type_](request)
```
thomascobb: If you do this on type(request) then you can put class identifiers in the break table

malcolm/core/request.py
```diff
@@ -56,7 +50,7 @@ def respond_with_error(self, message):
             message(str): Message explaining error
         """
 
-        response = Response.Error(self.id_, self.context, message=message)
+        response = Error(self.id_, self.context, message=message)
         self.response_queue.put(response)
 
     def respond_with_update(self, value):
```
thomascobb: This can go on Subscription

malcolm/core/request.py
```diff
@@ -66,7 +60,7 @@ def respond_with_update(self, value):
         Args:
             value (dict): Dictionary describing the new structure
         """
-        response = Response.Update(self.id_, self.context, value=value)
+        response = Update(self.id_, self.context, value=value)
         self.response_queue.put(response)
 
     def respond_with_delta(self, changes):
```
thomascobb: As can this

PR #98, opened 2016-07-15
Api docs

GDYendell:
Docs don't work properly because of import errors (cothread, tornado) most are fine except for importing CompoundGenerator in pointgeneratormeta, because that is in metas.\_\_init\_\_ so it affects a lot of things.


docs/dev/api.rst
```diff
@@ -0,0 +1,85 @@
+Malcolm API
```
thomascobb: Don't want to check in any generated files... If you create it as ./build/api.rst and make a relative path in index.rst and then it will already be there.

docs/dev/metas_api/blockmeta.rst
```diff
@@ -0,0 +1,7 @@
+BlockMeta
+---------
+
+.. module:: malcolm.metas.blockmeta
```
thomascobb: Can you make it so that the import path is as we end up using it. If you just make all the metas `.. module:: malcolm.metas` then this should work

docs/dev/statemachines_api/defaultstatemachine.rst
```diff
@@ -0,0 +1,7 @@
+DefaultStateMachine
```
thomascobb: Same with statemachines

malcolm/wscomms/wsclientcomms.py
```diff
@@ -1,6 +1,7 @@
 from collections import OrderedDict
 import json
 
+import tornado
```
thomascobb: can you put this in conf.py instead please?

PR #101, opened 2016-07-19
Update make publish to use https url explicitly

GDYendell:
Seems to build OK, not sure what the problem was.


GDYendell:
Makefile: Line 29: Why does it delete the .pypirc?

thomascobb:
Because I used to think it was a bad idea storing password in plain text... Don't any more

PR #102, opened 2016-07-21
Added all scalar capart subclasses

gilesknap:


malcolm/parts/ca/capart.py
```diff
@@ -8,9 +8,8 @@
 
 class CAPart(Part):
 
-    def __init__(self, name, process, block, meta, pv, rbv=None, rbv_suff=None,
-                 long_string=False):
-        super(CAPart, self).__init__(name=name, process=process, block=block)
+    def create_attribute(self, meta, pv, rbv=None, rbv_suff=None,
+                         long_string=False):
```
thomascobb: Don't need long string any more

tests/test_parts/test_ca/test_caschoicepart.py
```diff
@@ -0,0 +1,31 @@
+import os
```
thomascobb: tests/test_parts/test_ca/test_cachoicepart.py

PR #113, opened 2016-08-05
Ensure that when new blocks are added any subscriptions to the root get updates

ajgdls:
You probably want to check this change out @thomascobb.  It works for me, or at least it appears to work but the structure of the delta messages makes it difficult for me to be 100% confident, my eyes can't parse enough parentheses.


thomascobb:
Turns out that _handle_block_changes is doing the right thing anyway, I've made some significant changes in my branch anyway, so I'll add them there...

PR #124, opened 2016-08-15
Task changes

thomascobb:


tests/test_core/test_controller.py
```diff
@@ -41,5 +41,13 @@ def test_set_writeable_methods(self):
         self.c.register_method_writeable(m, "Ready")
         self.assertEqual(self.c.methods_writeable['Ready'][m], True)
 
+    def test_run_hook(self):
+        # TODO: write this
+        pass
+
+    def test_run_hook_raises(self):
+        # TODO: write this
+        pass
```
xen-0: Maybe we should write some tests?

thomascobb: That sounds like a good idea, are you volunteering? ;)

PR #139, opened 2016-09-16
Pva comms

ajgdls:
Implementation of pvput (client and server).  Correct decoding of variant unions.


malcolm/comms/pva/pvaclientcomms.py
```diff
@@ -28,25 +29,83 @@ def send_to_server(self, request):
...
-                return_object = Error(id_=request["id"], message="To be determined")
+                return_object = Error(id_=request["id"], message="PvAccess error")
 
             self.log_debug("Return object: %s", return_object)
             self.send_to_caller(return_object)
 
+        elif isinstance(request, Put):
+            self.log_debug("Put message with endpoint: %s", request["endpoint"])
+            self.log_debug("Put message with value: %s", request["value"])
+            try:
```
thomascobb: Can you put the contents of each try block into a separate function please

PR #156, opened 2016-10-13
First implementation of V4 monitors (server side).

ajgdls:


ajgdls:
@thomascobb coverage down :( but it will be resolved.

PR #162, opened 2016-10-28
Pva comms

ajgdls:
It's dirty, but time is short...


ajgdls:
@thomascobb don't merge this just yet, I might have time to improve coverage before the end of this week.

PR #209, opened 2017-05-15
Excalibur + VDSWrapperPart

GDYendell:


thomascobb:
Hmm, h5py is needed now, could you add a ci/install_hdf5.sh script and an entry in .travis.yml please?

thomascobb:
You can get the hdf5 libs like this:

`wget -O hdf5-src.tar.bz2 https://www.hdfgroup.org/package/bzip2/?wpdmdl=4300`

thomascobb:
Then unpack and build, and put on your path in some way...

thomascobb:
If you've got time...

GDYendell:
Oh I didn't see the further messages. I have done it the way I did it for vds-gen, which I believe I copied from the h5py repo.

Is this a suitable method?

thomascobb:
That's fine, but I'm guessing it won't support SWMR? Although we don't have any tests at the moment so maybe it doesn't matter. I'll merge this for now, could you think about how you would add tests for this that actually test the monitoring, reading and writing files please?

ulrikpedersen:
While there are no automatic SWMR tests in the system, this seems like the appropriate way to install HDF5 (always use OS packages whenever possible). Also it may not be possible to run SWMR testing on Travis-CI as there is a requirement for a proper POSIX compliant file-system - and Travis-CI may possibly be using NFS.

If you do want to install a later version of HDF5 the way to do it is
* Download (wget) the sources
* configure, build and install to a local location (--prefix)
* Add the install (--prefix) bin dir to the PATH
* Before installing the pip requirements; set the [HDF5_DIR and HDF5_VERSION](http://docs.h5py.org/en/latest/build.html#custom-installation) environment variables
* Setup Travis to [cache](https://docs.travis-ci.com/user/caching/) the installation dir (the --prefix dir) so that you don't need to rebuild HDF5 on every push

PR #216, opened 2017-10-02
Modify ADCore array counter attribute: split into read and write

callumforrester:
Implements functionality requested in http://jira.diamond.ac.uk/browse/BC-505
Merging this request closes http://jira.diamond.ac.uk/browse/BC-505

Note: ADCore and xmap both need to be modified as it turned out.


callumforrester:
I'm uncertain why the build is failing. Looking at the log, it doesn't seem to have anything to do with my changes.

PR #225, opened 2018-01-31
Added Support for ADAndor3 Module

callumforrester:
This pull request probably won't be merged!

Written as discussed with @thomascobb to showcase the changes needed to accommodate an ADAndor3 detector. These changes include:
- Everything from #222 
- Acquiring the readout time dynamically from a PV, as it can change.
- Compensating for the truncation in exposure time on the detector (restricted to the nearest multiple of the time it takes to read out a row of pixels, which itself varies depending on the model and configuration of the detector).

### DIFF
--- a/malcolm/modules/ADAndor3/blocks/__init__.py
+++ b/malcolm/modules/ADAndor3/blocks/__init__.py
@@ -1,0 +1,9 @@
+from malcolm.yamlutil import make_block_creator, check_yaml_names
+
+andor3_detector_driver_block = make_block_creator(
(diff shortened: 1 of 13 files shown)


callumforrester:
ee076e5 has greatly simplifed the problem by restricting the camera to global shutter mode. 
Unfortunately, this reduces the maximum framrate, so I will investigate support for rolling mode at a later date.

thomascobb:
This has now been superceded by Aaron's work which is now in master

PR #227, opened 2018-03-22
I14

EdWarrick:


malcolm/modules/pandablocks/parts/pandablocksfieldpart.py
```diff
@@ -40,4 +40,5 @@ def set_field(self, value):
         elif isinstance(self.meta, BooleanMeta):
             value = int(value)
         self.client.set_field(self.block_name, self.field_name, value)
+        # self.attr.set_value(value)
```
thomascobb: Can you remember why we commented this out?

malcolm/modules/excalibur/blocks/excalibur_detector_runnable_block.yaml
```diff
@@ -85,3 +85,12 @@
 
 - ADCore.parts.DatasetTablePart:
     name: DSET
+
+- excalibur.blocks.excalibur_munge_block:
```
thomascobb: Why is this needed? Is this just so you can disable the VDS writer and it still works?

PR #228, opened 2018-04-24
Fixed Excalibur VDS wrapper part and added unit test for it

MattTaylorDLS:


GDYendell:
As you're doing this now, you probably want 1-10-2, since that has VDS bugfixes.

MattTaylorDLS:
Yeah true. I wasn't sure whether it was better to keep in 'in check' with what we have in dls sw prod, as we don't have 1-10-2 there yet, but probably should.

GDYendell:
I am not sure why odin-data uses the --user flag, I have tried that with a virtualenv locally before and it doesn't work. Maybe travis has changed to use a venv and h5py hasn't been built in odin-data since.

MattTaylorDLS:
Yeah, might need to update it on there (if this works)

GDYendell:
Maybe try clearing the cache on travis?

MattTaylorDLS:
Hmm i don't know if that'll work. Does clearing the cache also remove the venv? I think the problem is that h5py is already installed on the venv. The cacheing i think is just for getting new py libraries if it needs to. If they're already installed, it doesn't need to install them again, whether from the cache or not

GDYendell:
But isn't the venv cached?

MattTaylorDLS:
I'm not sure. I cleared the cache on this pull request and it still fails to install h5py with the binaries as it already has h5py installed. Maybe if I clear the cache for the whole repo (maybe it's using the master cache?). Failing that, i'll just have to either use the --force-reinstall flag on the pip install, or 'manually' uninstall h5py and that should be ok

GDYendell:
I remember Alan and I having a similar problem and we cleared everything and added no-cache amongst other things maybe and it eventually worked OK.

MattTaylorDLS:
hmm well the python 2.7 builds are now passing, but the 3.5 and 3.6 aren't. It's failing at the same function, but not because it doesn't exist, but because of a Type error: expected bytes, str found.

PR #235, opened 2018-12-06
ADmerlin

btester271828:
Created block for merlin using ADCore Detector driver part
Also added attempted fix for git config issue in RHEL7 for managercontroller

### DIFF
--- a/malcolm/modules/ADMerlin/blocks/merlin_runnable_block.yaml
+++ b/malcolm/modules/ADMerlin/blocks/merlin_runnable_block.yaml
@@ -14,7 +14,7 @@
     value: |
       Device block corresponding to Merlin + stat + pos + hdf writer.
 
-      - Detector driver should have pv prefix $(pv_prefix):CAM
+      - Detector driver should have pv prefix $(pv_prefix):DET
       - Pos should have pv prefix $(pv_prefix):POS
       - Stat should have pv prefix $(pv_prefix):STAT
       - HDF should have pv prefix $(pv_prefix):HDF5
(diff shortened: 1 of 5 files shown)


malcolm/modules/ADMerlin/blocks/merlin_runnable_block.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+- builtin.parameters.string:
+    name: config_dir
+    description: Where to store saved configs
+
+- builtin.defines.docstring:
+    value: |
+      Device block corresponding to Merlin + stat + pos + hdf writer.
+
+      - Detector driver should have pv prefix $(pv_prefix):CAM
```
thomascobb: The block below suggests it should be :DET. Could you fix the docstring please?

malcolm/modules/builtin/controllers/managercontroller.py
```diff
@@ -125,12 +125,19 @@ def _run_git_cmd(self, *args):
...
             else:
                 self.log.debug("Git command completed: %s", output)
+                return output
 
     def do_init(self):
         super(ManagerController, self).do_init()
         # Try and make it a git repo, don't care if it fails
+        name = self._run_git_cmd("config", "--get", "user.name")
+        if name == "\n":
+            self._run_git_cmd("config", "user.name", "Malcolm")
```
thomascobb: Can we make it take the username and machine?

```suggestion
import getpass
import socket
            self._run_git_cmd("config", "user.email", "%s@%s" % (getpass.getuser(), socket.gethostname())
```

joeshannon:
Sorry I should have checked before but the git -c <config option> was only introduced in git 1.7.2. Rather annoying that RHEL6 system git is only just behind this at 1.7.1 so I think this might fail on RHEL 6.

PR #237, opened 2018-12-20
New DetectorDriver Part hardware triggering behaviour

gilesknap:
Now provide a list software_trigger_modes to ADCore.parts.DetectorDriverPart
defaults to 'Internal' only.
This list is compared against triggerMode state at configure time.

Replaces explicity passing 'hardware_triggered' flag.

### DIFF
--- a/malcolm/modules/ADCore/parts/__init__.py
+++ b/malcolm/modules/ADCore/parts/__init__.py
@@ -1,7 +1,7 @@
 from .datasetrunnablechildpart import DatasetRunnableChildPart
 from .datasettablepart import DatasetTablePart
 from .detectordriverpart import DetectorDriverPart, APartName, AMri, \
-    AMainDatasetUseful
+    AMainDatasetUseful, ASoftTriggerModes, USoftTriggerModes
 from .exposuredeadtimepart import ExposureDeadtimePart, AInitialAccuracy, \
     AInitialReadoutTime
 from .hdfwriterpart import HDFWriterPart, AFileDir, AFileTemplate, AFormatName
--- a/malcolm/modules/ADCore/parts/__init__.py
+++ b/malcolm/modules/ADCore/parts/__init__.py
@@ -1,7 +1,7 @@
 from .datasetrunnablechildpart import DatasetRunnableChildPart
 from .datasettablepart import DatasetTablePart
 from .detectordriverpart import DetectorDriverPart, APartName, AMri, \
-    AHardwareTriggered, AMainDatasetUseful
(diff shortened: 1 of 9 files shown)


malcolm/modules/ADCore/parts/__init__.py
```diff
@@ -1,7 +1,7 @@
 from .datasetrunnablechildpart import DatasetRunnableChildPart
 from .datasettablepart import DatasetTablePart
 from .detectordriverpart import DetectorDriverPart, APartName, AMri, \
-    AHardwareTriggered, AMainDatasetUseful
+    AMainDatasetUseful
```
thomascobb: Should export ASoftTriggerModes and USoftTriggerModes here

malcolm/modules/ADCore/parts/detectordriverpart.py
```diff
@@ -66,10 +68,14 @@ def configure(self,
                 generator.duration)
         self.actions.setup_detector(
             context, completed_steps, steps_to_do, **kwargs)
+        child = context.block_view(self.mri)
+        try:  # todo is this the best way to check if the block has triggerMode ??
+            self.is_hardware_triggered = child.triggerMode.value not in self.soft_trigger_modes
```
thomascobb: Do we need the try catch here? The child should be an areaDetector, so will always have triggerMode...

gilesknap: xmapdriverpart and reframeplugin both derive from detectordriverpart and do not have triggerMode. This try block means that these will always be 'hardware triggered' meaning that configure() will start them. 

I think that is correct behaviour for reframeplugin but not sure about xmap (not sure why it has no trigger mode?)

thomascobb: how about we only do this check if we have some soft trigger modes? Otherwise we default to hardware triggered. Then we would never do the check for xmap or the reframe plugin

gilesknap: Currently soft_trigger_modes defaults to ('Internal',). We would have to require that subclasses of detectordriverpart pass soft_trigger_modes=None if they do not have triggerMode (which is unintuitive). So we can change the default to None and require all detector types pass a soft_trigger_modes or continue to use the more direct approach of testing if triggerMode exists.

gilesknap: agreed on soft_trigger_modes defaults to None. committed changes.

PR #238, opened 2019-01-09
Add support for ADOdin

gilesknap:
To be added: support for pause/rewind, VDS and control of flush rate

### DIFF
--- a/malcolm/modules/ADOdin/__init__.py
+++ b/malcolm/modules/ADOdin/__init__.py
@@ -1,0 +1,1 @@
+from . import parts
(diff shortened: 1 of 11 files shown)


gilesknap:
I've added tests for ADOdin block and OdinWriterPart configure(). This is in keeping with other areadetector driver's test coverage. I'm looking at testing run() abort() etc. but need to work out how to do threading in tests. Experiments with cothread.Spawn are not working as yet.

gilesknap:
Add in ADOdin support

PR #242, opened 2019-02-08
Fix build

gilesknap:


thomascobb:
```python
        for attr, value in start_positions.items():
            child.when_value_matches(attr, value, timeout=1.0)
```

needs to be

```python
        for attr, value in sorted(start_positions.items()):
            child.when_value_matches(attr, value, timeout=1.0)
```

to avoid relying on unordered dict ordering

PR #243, opened 2019-02-21
Give Methods a "took" and "returned" log of the last call

thomascobb:
This will not be merged until the corresponding Malcolmjs changes have been made


thomascobb:
Looks like scanpointgenerator needs to use OrderedDict for the tests to pass...

GDYendell:
I think I wanted to use OrderedDict in scanpointgenerator at one point before, but we were worried about the performance hit. Why doesn't "==" work without converting to a string?

thomascobb:
Cos it's gone via a convert to PVA structure, then over the wire, then json_encode...

thomascobb:
This is a sign that the PVA structures are in undefined order, which will bite us at a later point if we don't fix it

thomascobb:
The performance hit is creating an OrderedDict in Point(), to_dict is called quite rarely so should be fine

thomascobb:
This has been merged into the malcolm4 branch as it contains minor breaking changes

PR #246, opened 2019-03-27
Excalibur vds

gilesknap:
Tests on python 3 are now passing.


gilesknap:
Hi have to leave now but have noticed that this has pulled in some of Toms changes and not sure why.

requirements/test.txt
```diff
@@ -10,6 +10,7 @@ plop==0.3.0
 pytest>=3.2.0
 pytest-cov<=2.6.0
 h5py>=2.8.0
-vdsgen>=0.3
-annotypes==0.9
+annotypes>=0.9
```
thomascobb:
```suggestion
annotypes==0.9
```
Need to leave this pinned as new versions of annotypes will break this...

malcolm/modules/ADOdin/parts/odinwriterpart.py
```diff
@@ -80,45 +238,36 @@ def configure(self,
                   completed_steps,  # type: scanning.hooks.ACompletedSteps
                   steps_to_do,  # type: scanning.hooks.AStepsToDo
                   generator,  # type: scanning.hooks.AGenerator
-                  fileDir='/tmp',  # type: AFileDir
-                  fileName="odin.hdf",  # type: AFileName
+                  fileDir,  # type: AFileDir
+                  formatName="odin",  # type: AFormatName
+                  fileTemplate="%s.hdf",  # type: AFileTemplate
```
thomascobb:
```suggestion
                  fileTemplate="%s.h5",  # type: AFileTemplate
```

setup.py
```diff
@@ -26,7 +26,7 @@ def get_version():
         raise RuntimeError('Unable to find version string in {0}.'
                            .format(VERSION_FILE))
 
-install_requires = ['numpy', 'scanpointgenerator>=1.5', 'annotypes']
+install_requires = ['numpy', 'scanpointgenerator>=1.5', 'annotypes', 'vdsgen']
```
thomascobb: Not install_requires. We need to define some optional requires targets for hdf5, pvaccess, etc. Cothread does need to be in this list now though...

tests/test_modules/test_ADOdin/test_odinwriterpart.py
```diff
@@ -1,58 +1,202 @@
...
+        assert detector_group.attrs['y_set_indices'] == '0'
+        assert detector_group.attrs['x_set_indices'] == '1'
+
+        # Check _set datasets
+        # N.B. units are encoded as ASCII in the original file, so come
+        # back as type byte in Python 3
+        stage1_x_set_dataset = vds_file['/entry/detector/x_set']
+        assert stage1_x_set_dataset[0] == 0
+        assert stage1_x_set_dataset[1] == 2
+        assert stage1_x_set_dataset.attrs['units'] == b'mm'
```
thomascobb: Is this really a bytes literal? Does h5py not convert to Python3 string?

thomascobb:
My commits are because of the re-rebase that we did, Github keeps track of things that git forget about...

tests/test_modules/test_ADOdin/test_odinwriterpart.py
gilesknap: It seems I lied about Python 3 tests working. In Python3 this is str but in Python2 it is bytes. Will change the test to case the value to str.

setup.py
gilesknap: I've removed vdsgen and added cothread. Are the optional targets a future TODO? The imalcolm entry point does not list any of the extras_requires entries, so at present setup.py does not seem to bring in any extra dependencies.

PR #256, opened 2019-06-20
Added stats block

btester271828:
Runs up a softIoc with some PVs like devIocStats provides, giving info on which version of pymalcolm is running, which machine is running it and where the config was loaded from.


malcolm/modules/stats/controllers/statscontroller.py
```diff
@@ -0,0 +1,126 @@
...
+            tags=[Widget.TEXTUPDATE.tag()]).create_attribute_model(
+            stats["kernel"])
+        self.field_registry.add_attribute_model("kernel", self.kernel)
+        self.pid = StringMeta(
+            "process ID of pymalcolm instance",
+            tags=[Widget.TEXTUPDATE.tag()]).create_attribute_model(
+            stats["pid"])
+        self.field_registry.add_attribute_model("pid", self.pid)
+
+        self.ioc = start_ioc(stats, prefix)
```
thomascobb: This should be called at InitHook

malcolm/modules/stats/controllers/statscontroller.py
```diff
@@ -0,0 +1,126 @@
...
+
+        self.register_hooked(HaltHook, self.ioc.terminate)
```
thomascobb: This will only work in a Part, so I think this should be turned into a Part...

thomascobb:
Superceded by #260

PR #258, opened 2019-07-02
Python37

gilesknap:
Changes to make it work on 3.7.
- new annotypes and other dependencies
- add 3.7 to travis
- use annotypes patched issubclass() in model.py


gilesknap:
Updated to use the suggested changes to annotypes.

PR #259, opened 2019-07-17
Velocity mode changes

gilesknap:
Tom,

I'm making a PR but this is a breaking change since Malcolm now requires trajectory prog v2 (which is checked in but not released).

Changes:

- fix velocity mode calculation for PREV_NEXT
- change turnaround points generation to use acceleration change points only and alway PREV_NEXT velocity mode
- error if the trajectory program is not v2


gilesknap:
This is superseded by #265

PR #260, opened 2019-07-23
Status screen

btester271828:
- Added malcom status block/part, showing yaml path, imalcolm path, os version etc.
- Added block for showing devIOCStats PVs, also parses configure/RELEASE to show dependencies in a table


malcolm/core/process.py
```diff
@@ -234,6 +234,25 @@ def add_controller(self, controller, timeout=None):
             if self.state == STARTED and should_publish:
                 self._publish_controllers(timeout)
 
+    def add_controllers(self, controllers, timeout=None):
```
thomascobb: Please can you make add_controller(controller) call add_controllers([controller])

malcolm/modules/stats/blocks/__init__.py
```diff
@@ -0,0 +1,5 @@
+from malcolm.yamlutil import make_block_creator, check_yaml_names
```
thomascobb: Can we rename the whole stats directory system please?

malcolm/modules/stats/parts/iociconpart.py
```diff
@@ -0,0 +1,38 @@
...
+    def update_icon(self, arch):
+        svg = self.initial_svg
+        if arch.ok:
+            if arch.startswith("WIND"):
+                # WIND River VXWorks
+                svg = os.path.split(__file__)[0] + "/../icons/vx_epics.svg"
+            elif arch.startswith("Linux"):
+                # Linux (typically RHEL)
+                svg = os.path.split(__file__)[0] + "/../icons/linux_epics.svg"
+            elif arch.startswith("Windows"):
```
thomascobb: Maybe put this first in the if statements just in case we ever report "WINDOWS" instead of "Windows"...

malcolm/core/tags.py
```diff
@@ -161,3 +162,8 @@ def method_return_unpacked():
     """
     tag = "method:return:unpacked"
     return tag
+
+def hidden_method():
```
thomascobb:
```suggestion
def method_hidden():
```
For consistency with the tag text

malcolm/modules/system/parts/iociconpart.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def update_icon(self, arch):
+        svg = self.initial_svg
+        if arch.ok:
+            if arch.upper().startswith("WINDOWS"):
+                svg = os.path.split(__file__)[0] + "/../icons/win_epics.svg"
+            elif arch.upper().startswith("LINUX"):
+                # Linux (typically RHEL)
+                svg = os.path.split(__file__)[0] + "/../icons/linux_epics.svg"
+            elif arch.upper().startswith("WIND"):
```
thomascobb: Should we check for WIND RIVER instead?

btester271828: The string returned by devIocStats is "WIND version 2.6"...maybe we could check for "WIND " (with the trailing space)?

PR #261, opened 2019-07-25
Sparse trajectories only

gilesknap:
This is OK to merge into master.

It implements the sparse PVT points when 'start of row' triggering is requested. It only works for linear rows but does not break non-linear rows (it just creates the same PVT points as before the change).

This branch is now separate from velocity-mode-changes and so it still works with the current version of the trajectory PROG.

### DIFF
--- a/malcolm/modules/pmac/parts/pmacchildpart.py
+++ b/malcolm/modules/pmac/parts/pmacchildpart.py
@@ -501,7 +501,7 @@ class PmacChildPart(builtin.parts.ChildPart):
         else:
             # otherwise skip this point if is is linear to previous point
             do_skip = next_point and points_are_joined and \
-                      self.is_equidistant(point, next_point)
+                      self.is_same_velocity(point, next_point)
 
         if do_skip:
             self.time_since_last_pvt += point.duration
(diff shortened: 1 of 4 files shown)


malcolm/modules/pmac/parts/pmacchildpart.py
```diff
@@ -570,3 +622,17 @@ def insert_gap(self, point, next_point, completed_steps):
...
         self.profile["userPrograms"][-1] = user_program
+
+    def is_equidistant(self, p1, p2):
+        result = False
+        # handle p1 = None, i.e. no previous point
+        if p2.duration == p2.duration:
+            result = True
+            for axis_name, motor_info in self.axis_mapping.items():
+                if not np.isclose(
+                    p1.lower[axis_name] - p1.positions[axis_name],
```
thomascobb: Should we alos check upper-positions?

malcolm/modules/pmac/parts/pmacchildpart.py
```diff
@@ -570,3 +622,17 @@ def insert_gap(self, point, next_point, completed_steps):
         self.profile["velocityMode"][-1] = CURRENT_TO_NEXT
         user_program = self.get_user_program(PointType.START_OF_ROW)
         self.profile["userPrograms"][-1] = user_program
+
+    def is_equidistant(self, p1, p2):
```
thomascobb: should this be is_same_velocity?

malcolm/modules/pmac/parts/pmacchildpart.py
gilesknap: yep

malcolm/modules/pmac/parts/pmacchildpart.py
gilesknap: yes

PR #265, opened 2019-09-30
New Velocity Profile approach with 2 points only - with quantization of turnaround times.

gilesknap:
Tom,
I'm not sure if you want this in master yet?
But I'd like you to take a look at it please.

The unit tests are passing and also my system tests work with this (the one that shows the need for quantizing the times still breaks)

However, test_configure_continuous in testPCompPart is failing and I do not understand it. I can investigate this tomorrow.


gilesknap:
I just realized that this PR brought in both the sparse trajectories commits too. I guess that means we did not have them in master yet?

gilesknap:
I think the sparse trajectories commits are a red herring because I cherry picked them over already when disassociating 'spare trajectories' and 'velocity mode changes'.

So this PR does include the 'velocity mode changes' which are a necessary part. This PR is therefore a superset of #259. Again this PR contains a breaking change that requires a new Trajectory Scan PROG 1 (which is already on P45).

gilesknap:
#cant_sleep_with_this_problem_unsolved

@thomascobb. Quantization is in and working. I still need to put back in settle time but have a pretty simple solution for that (it will go into the VelocityProfile class itself).

I would also like your advice as to why test_configure_continuous is failing in test_ADPandABlocks/test_pandaseqtriggerpart.py

Thanks,
giles.

gilesknap:
@thomascobb settle time added

gilesknap:
I think the quantize will do this to a snake turnaround. https://photos.app.goo.gl/MH77J2zexAV9KVpL6. I think that is OK, but need to do some good system tests.

gilesknap:
ooops - working in Python 3 caught me out. I need to fix the type hints.

gilesknap:
After discussion with @thomascobb I have realized that we can preserve symmetry when  quantizing by using the following algorithm:
```
        # round up the two acceleration times t1, t2 to the nearest millisecond
        # then round up the middle constant velocity tm such that the total
        # time tv2 is tv2 + 2 milliseconds rounded up
        # This approach preserves symmetry but at the same time ensures that
        # total time is increased deterministically plus acceleration and
        # vm are decreased
```
This has been applied and the system tests now show turnarounds with zero velocity their middle section.

gilesknap:
@thomascobb 

I believe this is ready to merge

- all checks are now passing
- symmetrical quantization is in
- the system tests are also working, including those that failed due to points being too close

malcolm/modules/pmac/__init__.py
```diff
@@ -1 +1,2 @@
 from . import parts, infos, util
+from .velocityprofile import VelocityProfile
```
thomascobb: Could you put VelocityProfile in util please?

malcolm/modules/pmac/parts/pmactrajectorypart.py
```diff
@@ -1,11 +1,14 @@
 import numpy as np
 from annotypes import add_call_types, Anno, Array
 
-from malcolm.core import PartRegistrar, BooleanMeta, config_tag, Widget, \
-    NumberMeta
+from malcolm.core import PartRegistrar, Widget, \
```
thomascobb: Can you put IncompatibleError in malcolm.core.__init__ and import it from there please?

malcolm/modules/pmac/velocityprofile.py
```diff
@@ -0,0 +1,451 @@
...
+        Returns bool: true if this profile requires quantization:
+        """
+
+        def is_quantized(val  # type: float
+                         ):
+            decimals = val % 1 * 1000
+            return np.isclose(decimals, np.round(decimals))
+
+        times = np.array([self.t1, self.tm, self.t2])
+        result = not np.vectorize(is_quantized)(times).all()
```
thomascobb: Can you write this as:
```python
decimals = (times % 1) * 1000
result = not np.isclose(decimals, np.round(decimals)).all()
```

malcolm/modules/pmac/velocityprofile.py
```diff
@@ -0,0 +1,451 @@
...
+            time_array.insert(2, self.t1 + self.tm)
+            velocity_array.insert(2, self.vm)
+        if self.settle_time > 0:
+            time_array.append(time_array[-1] + self.settle_time)
+            velocity_array.append(self.v2)
+        time_array = np.around(time_array, 10)
+
+        # some of the math results in tiny fractions which affect some of the
+        # tests - round to 10 decimals
+        velocity_array = np.around(velocity_array, 10)
```
thomascobb: Is this actually safe to do? We never store the absolute position, so rounding the velocities and times might change the end position...

tests/test_modules/test_pmac/test_motorinfo.py
```diff
@@ -18,16 +20,49 @@ def setUp(self):
...
     def test_turnaround(self):
         # 0_| \
         #   |  \
         v1 = 0.1
         v2 = -0.1
         distance = 0.0
         time_array, velocity_array = self.o.make_velocity_profile(
+            v1, v2, distance, 0.0).make_arrays()
+        self.check_distance(distance, time_array, velocity_array)
+        assert np.isclose(time_array, [0.0, 0.05, 0.1]).all()
```
thomascobb:
```suggestion
        assert time_array == pytest.approx([0.0, 0.05, 0.1])
```

Maybe a bit more readable? Debatable though...

tests/test_modules/test_pmac/test_pmacchildpart.py
```diff
@@ -96,81 +93,150 @@ def test_validate(self):
...
+                           -0.13748751]),
+                      b=pytest.approx(
+                          [0., 0., 0., 0., 0., 0., 0., 0.01231466, 0.04975248,
+                           0.05024752, 0.08768534, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1,
+                           0.1, 0.1]),
+                      csPort='CS1',
+                      timeArray=pytest.approx(
+                          [99950, 500250, 500250, 500250, 500250, 500250,
+                           500250, 100000, 101000, 1000, 101000, 100000, 500250,
+                           500250, 500250, 500250, 500250, 500250, 99950]),
```
thomascobb: It's a shame we need this 1ms point in the middle of the turnaround, but I guess it's a by-product of the way we calculate the velocity

malcolm/modules/pmac/velocityprofile.py
gilesknap: I think only by approx 10**-10. Is that an issue?

tests/test_modules/test_pmac/test_pmacchildpart.py
gilesknap: Its a product of the way we extend the times in order to quantize. If there was no middle section then the 0 length section gets extended to 1 ms.

I think we could use a different quantize approach for 'pointy hats'. Initially, I did have this but it broke because if there was a mix of pointy and flat hats then the times got extended to different values. Having re-thought the time extension to preserve symmetry, I now believe I could adjust it to work for pointy hats too.

tests/test_modules/test_pmac/test_motorinfo.py
gilesknap: it looks a bit nicer

malcolm/modules/pmac/__init__.py
gilesknap: I see that you have a strict structure, but feel this makes a complex bit of code a little less legible.

Precedent: pandablocksclient.py ?

malcolm/modules/pmac/parts/pmactrajectorypart.py
gilesknap: Done

malcolm/modules/pmac/velocityprofile.py
gilesknap: Doh! that's what I meant to say :-)

malcolm/modules/pmac/velocityprofile.py
gilesknap: I changed it to 12 decimals

tests/test_modules/test_pmac/test_motorinfo.py
gilesknap: Done

tests/test_modules/test_pmac/test_pmacchildpart.py
gilesknap: Modified the quantize to preserve pointy hats. See the comment in VelocityProfile.quantize()

gilesknap:
@thomascobb I have committed your suggested changes, but have left 2 conversations outstanding. Please take a look at the outstanding conversations. Thanks.

PR #267, opened 2019-10-09
Odin VDS dataset fixes

btester271828:
made dataset names for uid, sum & secondary set to link parameters passed to OdinWriterPart

### DIFF
--- a/malcolm/modules/ADOdin/blocks/odin_runnable_block.yaml
+++ b/malcolm/modules/ADOdin/blocks/odin_runnable_block.yaml
@@ -30,6 +30,22 @@
     description: Design to load at init
     default:
 
+- builtin.parameters.string:
+    name: sum_name
+    description: Name of sum dataset
+    default: SUM
+
+- builtin.parameters.string:
+    name: uid_name
+    description: Name of uid dataset
+    default: UID
+
+- builtin.parameters.string:
+    name: secondary_set
+    description: Name of secondary dataset to link in nxs file
+    default: sum
+
+
 - builtin.defines.docstring:
     value: |
       Device block corresponding to ADOdin + Odin file writer plugin .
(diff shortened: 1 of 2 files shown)


malcolm/modules/ADOdin/blocks/odin_runnable_block.yaml
```diff
@@ -30,6 +30,22 @@
...
 
+- builtin.parameters.string:
+    name: sum_name
+    description: Name of sum dataset
+    default: SUM
+
+- builtin.parameters.string:
+    name: uid_name
+    description: Name of uid dataset
+    default: UID
```
MattTaylorDLS: I'd suggest making the default uid instead of UID.

malcolm/modules/ADOdin/blocks/odin_runnable_block.yaml
```diff
@@ -30,6 +30,22 @@
     description: Design to load at init
     default:
 
+- builtin.parameters.string:
+    name: sum_name
+    description: Name of sum dataset
+    default: SUM
```
MattTaylorDLS: I'd suggest making the default sum instead of SUM.

malcolm/modules/ADOdin/parts/odinwriterpart.py
```diff
@@ -235,6 +243,20 @@ class OdinWriterPart(builtin.parts.ChildPart):
     layout_filename = None  # type: str
     exposure_time = None  # type: float
 
+    def __init__(self,
+                 name,  # type: APartName
+                 mri,  # type: AMri
+                 initial_visibility=True,  # type: AInitialVisibility
+                 uid_name="UID",  # type: AUidName
+                 sum_name="SUM",  # type: ASumName
```
MattTaylorDLS: Should these be lower case as well?

PR #269, opened 2019-10-11
Bs new

dinojugosloven:
A new part in PMAC that enables sending a demand to a motor at configure step.


malcolm/modules/pmac/parts/motorpremovepart.py
```diff
@@ -0,0 +1,40 @@
+from annotypes import Anno, add_call_types
+
+from malcolm.core import PartRegistrar
+from malcolm.modules import builtin, scanning
+
+APartName = builtin.parts.APartName
+AMri = builtin.parts.AMri
+AIV = builtin.parts.AInitialVisibility
```
thomascobb: We need to maintain the same name for the annotypes
```suggestion
AInitialVisibility = builtin.parts.AInitialVisibility
```

tests/test_modules/test_pmac/test_bs_manager_block.yaml
```diff
@@ -0,0 +1,16 @@
+- builtin.parameters.string:
```
thomascobb: We don't need this file anymore

PR #271, opened 2019-10-21
Feature/shutterpart

BenBradnick:
A generic shutter part which can controls opening and closing shutter through the CAChoicePart (assumes the shutter has some sort of positioner with two relevant positions to move between).


BenBradnick:
Hi Tom, I believe it is now designed as required.

PR #272, opened 2019-10-25
Logging changes for Syslog to Graylog

MattTaylorDLS:
Changes for automatic forwarding of logging from syslog to graylog.

N.B. The current logging config in imalcolm will mean graylog may receive two messages for the same log message, one from gelf and one from syslog, on any machine where syslogging forwarding from LOCAL0 has been enabled.
When this change has been rolled out to all controls machines, the gelf logging can be turned off, and the config removed from imalcolm.

### DIFF
--- a/malcolm/syslogger.py
+++ b/malcolm/syslogger.py
@@ -21,9 +21,8 @@ class JsonSysLogHandler(logging.handlers.SysLogHandler):
         Emit a record adding the extra fields as a json string,
         removing any unwanted fields first.
         """
-        extra = record.__dict__.copy()
-        for field_to_remove in self.fields_to_remove:
-            extra.pop(field_to_remove, None)
+        extra = {k: v for k, v in record.__dict__.items()
+                 if k not in self.fields_to_remove}
 
         record.extra = json.dumps(extra)
         super(JsonSysLogHandler, self).emit(record)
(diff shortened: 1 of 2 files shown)


malcolm/syslogger.py
```diff
@@ -0,0 +1,29 @@
...
+                        'created',
+                        'threadName',
+                        'msecs']
+
+    def emit(self, record):
+        """
+        Emit a record adding the extra fields as a json string,
+        removing any unwanted fields first.
+        """
+        extra = record.__dict__.copy()
```
thomascobb:
```suggestion
        extra = {k: v for k, v in record.__dict__.items()
                 if k not in self.fields_to_remove}
```

tests/test_core/test_loggable.py
```diff
@@ -1,5 +1,4 @@
 import unittest
-from mock import patch
```
thomascobb: Don't we still need this?

MattTaylorDLS: Oops a bit over zealous in my deleting after I'd added some more imports from mock I didn't need. Yes we still need it, so have put back in

malcolm/syslogger.py
MattTaylorDLS: Changed

PR #274, opened 2019-11-04
Quantization fix

gilesknap:
This adds configuration of the quantization interval to scanning.parts.MinTurnaroundPart.

However, two of the tests are failing on Travis but not locally. Have you seen this issue? :
```
_______________________ TestPMACChildPart.test_step_scan _______________________
Traceback (most recent call last):
  File "/home/travis/build/dls-controls/pymalcolm/tests/test_modules/test_pmac/test_pmacchildpart.py", line 517, in test_step_scan
    assert self.child.handled_requests.mock_calls[-1].kwargs['a'] == \
TypeError: tuple indices must be integers or slices, not str
```

### DIFF
--- a/malcolm/modules/pmac/velocityprofile.py
+++ b/malcolm/modules/pmac/velocityprofile.py
@@ -1,5 +1,6 @@
-import numpy as np
 from math import sqrt, fabs
+
+import numpy as np
 
 HAT = 1
 INVERSE_HAT = 2
@@ -48,6 +49,7 @@ class VelocityProfile:
         t2: the time interval from vm to v2
         vm: the velocity in the middle portion of the hat or ramp
         t_out: the adjust time if t is too short to achieve d
+        interval: minimum interval between any two points in a profile
 
         d_trough: minimum distance possible between v1,v2 in tv2
         d_peak: maximum distance possible between v1,v2 in t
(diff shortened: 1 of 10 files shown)


gilesknap:
@thomascobb sub quantization turnaround is now added, with a test.
All tests are passing.

malcolm/modules/pmac/velocityprofile.py
```diff
@@ -64,7 +65,8 @@ def __init__(
             tv2,  # type: float
             a,  # type: float
             v_max,  # type: float
-            settle_time=0  # type: float
+            settle_time=0,  # type: float
+            interval=0.002  # type: float
```
thomascobb: Repeating 0.002 seems wrong. Should this be MIN_INTERVAL? Or should there be no default?

gilesknap: I have changed the default to be 0 and for that to mean no quantization occurs.
Thus the default reverts to behaviour before the new parameter was added.

thomascobb:
Just to check (can't do the math for the test in my head), if a turnaround takes strictly less than MIN_INTERVAL, then the result is two points exactly MIN_INTERVAL apart.

gilesknap:
Yes you just get a start and end point. See the test at the end of https://github.com/dls-controls/pymalcolm/blob/quantization-fix/tests/test_modules/test_pmac/test_pmacchildpart.py
called test_minimum_profile to verify that it does what you expect.

PR #278, opened 2019-12-02
Fix pandablocksclient to work with python3

hir12111:


malcolm/__init__.py
```diff
@@ -4,8 +4,12 @@
...
+    # In a release there will be a static version file written by setup.py
 except ImportError:
-    # Otherwise get the release number from git describe
-    from ._version_git import __version__
+    try:
+        # Otherwise get the release number from git describe
+        from ._version_git import __version__
+    except Exception:
+        # Oh well.
+        __version__ = 'undefined'
```
thomascobb: What exception did you get? We need to fix it in _version_git rather than swallowing it here...

hir12111: OSError: [Errno 2] No such file or directory
Caused by the missing git executable in the target.

thomascobb: Ok, updated versiongit in master. Could you remove changes in this file and rebase onto master please?

PR #280, opened 2019-12-06
Minor corrections to tutorials

thomascobb:


docs/tutorials/pmac.rst
```diff
@@ -17,8 +17,7 @@ We assume for this tutorial that you have created one or more IOCS that contain:
 - A GeoBrick Controller template from the `EPICS pmac`_ module with PV prefix
   that looks like ``BLxxI-MO-BRICK-01``.
 - A GeoBrick Trajectory template with the same prefix.
-- One or more CS templates.
-- One or more dls_pmac_asyn_motor or dls_pmac_cs_asyn_motor instances.
+- One or more dls_pmac_asyn_motor instances.
```
thomascobb: We still need the CS templates, all co-ordinated motion is done via CS templates, even if we only move straight through mapped motors

PR #281, opened 2019-12-11
Feature/automated testing

BenBradnick:
This adds a ScanRunnerPart for testing a scan block. Scan parameters are first loaded from a user-written YAML file, and then the part can be ran to go through each set of parameters. The part writes scan data and a small report file inside a defined top-level output directory.


BenBradnick:
I combined your suggestion into a single object as the generator contains all of the physical parameters.

Let me know if there is anything else I should modify - otherwise I believe it is ready.

malcolm/modules/scanning/parts/scanrunnerpart.py
```diff
@@ -0,0 +1,498 @@
...
+    def get_file_contents(self):
+        # type: () -> str
+
+        try:
+            with open(self.scan_file.value, "r") as input_file:
+                return input_file.read()
+        except IOError:
+            self.set_runner_state(RunnerStates.FAULT)
+            self.runner_status_message.set_value("Could not read scan file")
+            raise IOError("Could not read scan file")
```
thomascobb:
```suggestion
            raise
```
I would just raise here, as the error message from open() is probably more descriptive and the traceback is useful to have

malcolm/modules/scanning/parts/scanrunnerpart.py
```diff
@@ -0,0 +1,498 @@
...
+
+    def parse_yaml(self, string):
+        # type: (str) -> ...
+        try:
+            parsed_yaml = yaml.safe_load(string)
+            return parsed_yaml
+        except yaml.YAMLError:
+            self.set_runner_state(RunnerStates.FAULT)
+            self.runner_status_message.set_value("Could not parse scan file")
+            raise yaml.YAMLError("Could not parse scan file")
```
thomascobb: Possibly ditto? Would be useful to have the traceback

malcolm/modules/scanning/parts/scanrunnerpart.py
```diff
@@ -0,0 +1,498 @@
...
+
+        return LineGenerator(axes, units, start, stop, size, **kwargs)
+
+    def parse_compound_generator(self, entry):
+        # type: (dict) -> CompoundGenerator
+        duration = entry['duration']
+        generators = []
+        generators_dict = entry['generators']
+        for generator in generators_dict:
+            generators.append(self.parse_line_generator(generator['line']))
```
thomascobb:
```suggestion
            generators.append(LineGenerator.from_dict(generator['line']))
```
Should work...

malcolm/modules/scanning/parts/scanrunnerpart.py
```diff
@@ -0,0 +1,498 @@
...
+    def create_and_get_sub_directory(self, root_directory):
+        # type: (str) -> str
+        today_str = self.get_current_datetime(time_separator="-")
+        sub_directory = "{root}/{scan_mri}-{date}".format(
+            root=root_directory, scan_mri=self.mri, date=today_str)
+        self.create_directory(sub_directory)
+        return sub_directory
+
+    @add_call_types
+    def run(self, context):
```
thomascobb: I take it we don't care about aborting these scans when they are kicked off?

malcolm/modules/scanning/parts/scanrunnerpart.py
```diff
@@ -0,0 +1,498 @@
...
+                set_name=set_name, scan_no=scan_number
+            ))
+
+        # Make individual scan directory
+        scan_directory = self.create_and_get_scan_directory(
+            set_directory, scan_number)
+
+        # Check if scan can be reset or run
+        while scan_block.state.value is RunnableStates.ABORTING:
+            cothread.Sleep(0.1)
```
thomascobb: Would be better to do context.sleep

malcolm/modules/scanning/parts/scanrunnerpart.py
BenBradnick: I have changed this accordingly.

malcolm/modules/scanning/parts/scanrunnerpart.py
BenBradnick: This has also been changed.

malcolm/modules/scanning/parts/scanrunnerpart.py
BenBradnick: It did work, so I've committed the change.

malcolm/modules/scanning/parts/scanrunnerpart.py
BenBradnick: This is now using context.sleep.

malcolm/modules/scanning/parts/scanrunnerpart.py
```diff
@@ -0,0 +1,487 @@
...
+        duration = entry['duration']
+        generators = []
+        generators_dict = entry['generators']
+        for generator in generators_dict:
+            generators.append(LineGenerator.from_dict(generator['line']))
+
+        kwargs_list = ['continuous', 'delay_after']
+        kwargs = self.get_kwargs_from_dict(entry, kwargs_list)
+
+        return CompoundGenerator(generators, duration=duration, **kwargs)
```
thomascobb: You should be able to do CompoundGenerator.from_dict(entry) here too, as long you have set entry["generators"] = generators

malcolm/modules/scanning/parts/scanrunnerpart.py
BenBradnick: I was planning on adding it at some point. What would the best way of handling this be?

malcolm/modules/scanning/parts/scanrunnerpart.py
BenBradnick: I have made this change and added a check for a negative generator duration.

malcolm/modules/scanning/parts/scanrunnerpart.py
thomascobb: Store context from run, then make a new function abort() that calls context.stop(). That will catch run() when it is waiting for the scan to run, but not abort the current running scan. To do that you would also have to abort() the last scan that was run.

BenBradnick:
I have added abort which first stops the stored context and then aborts the current scan (I believe this is the correct order?).

More tests were also added to improve coverage.

BenBradnick:
I believe this is now at the point ready for merging, if you are happy @thomascobb

PR #283, opened 2019-12-12
Feature/beamselector

dinojugosloven:
New beam selector part (specific for K11)

### DIFF
--- a/malcolm/modules/pmac/parts/beamselectorpart.py
+++ b/malcolm/modules/pmac/parts/beamselectorpart.py
@@ -42,7 +42,7 @@ class BeamSelectorPart(PmacChildPart):
             self.tomoAngle = float(tomoAngle)
             self.diffAngle = float(diffAngle)
             self.move_time = float(moveTime)
-        except:
+        except ValueError:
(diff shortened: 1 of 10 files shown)


malcolm/modules/pmac/parts/beamselectorpart.py
```diff
@@ -0,0 +1,138 @@
+from malcolm.modules import builtin, scanning
```
thomascobb:
```suggestion
# Treat all division as float division even in python2
from __future__ import division

from malcolm.modules import builtin, scanning
```

tests/test_modules/test_pmac/test_beamselectorpart.py
```diff
@@ -0,0 +1,145 @@
+
```
thomascobb:
```suggestion
# Treat all division as float division even in python2
from __future__ import division

```

malcolm/modules/pmac/parts/beamselectorpart.py
dinojugosloven: Thanks Tom. I am working on this now.

PR #284, opened 2019-12-17
Fix/andor driver part exposure kwargs

BenBradnick:
This fixes the issue with duplicate exposures in kwargs being passed from setup_detector in AndorDriverPart. I have also refactored the method and added some tests.
I also noticed the type hint for duration in DetectorDriverPart was int when I believe it should be a float.


thomascobb:
when codacy warnings are fixed this looks fine

PR #287, opened 2020-01-13
Add Trajectory Scan Triggering documentation

MJGaughran:
This was adapted from internal documentation written by Giles Knap.


docs/reference/triggering.rst
```diff
@@ -0,0 +1,88 @@
...
+corresponding trigger signals.
+
+The graph at the bottom shows the states of the GPIO lines as the PVT scan
+progresses through its points.
+
+Note that the acquisition time is shorter than the live time. This trigger
+signal for the detector is controlled by the PandA. At present it pulses after
+the transition of the live signal, delayed by half the readout time. The end of
+acquisition is determined by the settings on the detector and this would be set
+to live time - readout time.
```
thomascobb: There is nothing automatic that does this at the moment, although you can hard code it with a pulse Block.
When we roll out the pulse trigger part then it will do that automatically, and make the length of the trigger the same length as the exposure time

MJGaughran: After speaking to Giles about this, as far as I can tell the paragraph does correctly describe the trigger signal behaviour for current systems at Diamond. Do you have an example where we don't handle the readout time in such a manner?

Even if it is only a matter of convention, I still think this documentation should reflect current usage. There is a lot of flexibility when configuring Malcolm and PandA, and detailing it here could complicate things.

MJGaughran: Tom has suggested (in person) that this readout time delay is typically ignored for the majority of detectors.
I will amend the documentation to reflect this.

PR #290, opened 2020-01-22
Feature/breakpoints

dinojugosloven:
Changes to RunnableController to accept a list of breakpoints at which it will return the control to the operator during a scan.


malcolm/modules/scanning/controllers/runnablecontroller.py
```diff
@@ -35,9 +35,9 @@
 ATemplateDesigns = builtin.controllers.ATemplateDesigns
 
 
-def get_steps_per_run(generator, axes_to_move):
-    # type: (CompoundGenerator, List[str]) -> int
-    steps = 1
+def get_steps_per_run(generator, axes_to_move, breakpoints):
+    # type: (CompoundGenerator, List[str], List[int]) -> int
```
thomascobb: Please update the type hints to match. This now returns List[int]

malcolm/modules/scanning/controllers/runnablecontroller.py
```diff
@@ -50,9 +50,12 @@ def get_steps_per_run(generator, axes_to_move):
                 "Axis %s is not in %s" % (axis, axes_to_move)
             axes_set.remove(axis)
         # Now multiply by the dimensions to get the number of steps
-        steps *= dim.size
-    return steps
+        steps[0] *= dim.size
+
+    if len(breakpoints) > 1:
+        steps = breakpoints
```
thomascobb: We should validate that the sum of the breakpoints is less than or equal to steps[0]

tests/test_modules/test_scanning/test_runnablecontroller.py
```diff
@@ -443,3 +444,279 @@ def test_error_in_pause_returns_run(self):
         with self.assertRaises(AbortedError):
             f.result()
 
+    def test_tomography(self):
```
thomascobb: Please add a test [Line(y, size=2), Line(x)] with axesToMove=[x] and 3 breakpoints. It should break at each breakpoint, then where y steps, then at each breakpoint again.

malcolm/modules/scanning/controllers/runnablecontroller.py
```diff
@@ -466,7 +470,15 @@ def do_run(self, hook):
         self.abortable_transition(ss.POSTRUN)
         completed_steps = self.configured_steps.value
         if completed_steps < self.total_steps.value:
-            steps_to_do = self.steps_per_run
+            if len(self.steps_per_run) > 1:
+                if self._breakpoint < len(self.steps_per_run):
+                    steps_to_do = self.steps_per_run[self._breakpoint]
+                    self._breakpoint += 1
```
thomascobb: You increment it here, but you never set it back to zero (or one)

malcolm/modules/scanning/controllers/runnablecontroller.py
```diff
@@ -466,7 +470,15 @@ def do_run(self, hook):
...
+                # handle a case when the endpoint is not provided
+                else:
```
thomascobb: What happens in this case if we aren't being asked to move all axes?

malcolm/modules/scanning/controllers/runnablecontroller.py
dinojugosloven: I added an initialization line for the index in on_configure() function.
Additionally, it can be also rewound in pause().

tests/test_modules/test_scanning/test_runnablecontroller.py
dinojugosloven: What you are asking is basically a mapping scan. If I recall correctly, we specified at the beginning that the breakpoints will be used only for tomography.

dinojugosloven: I can still check it, of course.

thomascobb: We have added a feature, we need to make sure it works in combination with all the other sorts of arguments you would pass to the function, even if we don't have a use case for that at the moment

dinojugosloven: You mean it should fail safely, I see.

thomascobb:
Can we have a unit test with a grid scan, when Malcolm does only x, and 2 breakpoints in the line. Also, can we have one with the same grid scan, but where the sum of the breakpoints is greater than the number of Malcolm executed points in an inner scan, Malcolm should error (the existing code just warns).

BenBradnick:
> 
> 
> Can we have a unit test with a grid scan, when Malcolm does only x, and 2 breakpoints in the line. Also, can we have one with the same grid scan, but where the sum of the breakpoints is greater than the number of Malcolm executed points in an inner scan, Malcolm should error (the existing code just warns).

Sure, I'll look into adding these tests. I will also check the status of the previous conversations.

malcolm/modules/scanning/controllers/runnablecontroller.py
BenBradnick: The index does get modified as Dino's comment above.

tests/test_modules/test_scanning/test_runnablecontroller.py
BenBradnick: I have added several tests and refactored the code to handle the outer axes accordingly:

- 1D inner scan with 1 outer axis that Malcolm does not move
- 1D inner scan with 2 outer axes that Malcolm does not move
- 2D inner scan with 1 outer axis that Malcolm does not move

I believe the code should handle outer axes now. Although I'm unsure if having a single Dimension with 2 axes where 1 is moved by Malcolm and 1 would be valid - as this would affect how I calculate the number of outer steps (where I multiply the inner steps by the size of each outer dimension).

malcolm/modules/scanning/controllers/runnablecontroller.py
BenBradnick: This is now refactored. Now if the last breakpoint doesn't finish the scan an additional breakpoint is added to complete the inner scan. This is done before multiplying the list of breakpoints for each outer step.

BenBradnick:
I have made some modifications to this branch as noted in the comments above:

- Should now supported multiple dimensions of outer axes in the scan
- An exception is now raised at configure time if the sum of steps in between each breakpoint exceeds the total number of steps of the inner scan
- Tests have been added to test both of these additions

BenBradnick:
@thomascobb I believe this is ready for your eyes again.

malcolm/modules/scanning/controllers/runnablecontroller.py
```diff
@@ -636,18 +712,41 @@ def pause(self, lastGoodStep: ALastGoodStep = -1) -> None:
...
-        self.configured_steps.set_value(completed_steps + steps_to_do)
+
+        if self.use_breakpoints:
+            self.breakpoint_index -= 1
+            # offset = sum(self.steps_per_run[0:self.iBreakpoint])
+            # assert offset < completed_steps
+            in_run_steps = (
+                completed_steps % self.breakpoint_steps[self.breakpoint_index]
+            )
+            steps_to_do = self.breakpoint_steps[self.breakpoint_index] - in_run_steps
```
thomascobb, line 728: Repeated code between these two branches, can we move it outside the if statement please?

thomascobb (changes requested):

BenBradnick:
@thomascobb duplicated code has now been moved. Are other changes required?

thomascobb (approved):

PR #302, opened 2020-03-04
Fix bc 1255

gilesknap:
Merging in two bug fixes:

- BC-1255 add stretching of turnaround points that are > 4s
- I08-232 spiral step scans causing CS runtime errors. This was a bug in the total time calculation for a turnaround that occurred for some very specific turnaround distances.

ALSO restored the loss of performance in a sparse trajectory due to dumb indexing of delay_after.


thomascobb:
As discussed in person, the array version of points_joined needs to consider delay_after, and the single point version of points_joined needs deleting

gilesknap:
@thomascobb I have added a check for delay_after into pmac.util.all_points_joined(). 

This is the correct place for it, previously I had it in pmac.parts.pmacchildpart.calculate_generator_profile.

I have not added a test because the behaviour is checked adequately in test_beamselectorpart.py.

However, what would have been broken and is not tested is PandaSeqTriggerPart which also uses 
all_points_joined.

@MJGaughran please can you add a test to test_pandaseqtriggerpart.py that checks this? Come chat to me for details.

MJGaughran:
At the moment, delay_after is inserted into a given generator by BeamSelectorPart, for a generator of only one point. BeamSelectorPart also modifies the generation of trigger signals, confusing things further.

After discussing this issue, we have decided that we need explicit tests for delay_after for both pmac and panda parts.

gilesknap:
@MJGaughran I have added a test for delay_after. I have also snuck in a fix that honours minimum interval for the first point in a scan.

Please can you add a delay_after test for PandaSeqTriggerPart. Thanks.

gilesknap:
@MJGaughran I forgot to mention that it may not be obvious that although delay_after has a value for every point it the generator it is currently initialized in the CompundGenerator constructor as the same value for every point in the scan. Feel free to call 07505 004 505 if this is not clear.

MJGaughran:
I fixed the calculation for the blind time calculation to take the maximum of the delay_after and min_turnaround, in the same way as pmacchildpart.

The test I added also uses the same generator that is used in pmacchildpart.

MJGaughran:
@gilesknap could you please resolve the conflicts on this branch?

gilesknap:
OK looking at this now.

gilesknap:
OK this is turning out to be challenging. Signing off for the day. Hope to have it done before you start work tomorrow :-)

MJGaughran:
Do we think this can be merged now? @gilesknap @thomascobb

gilesknap:
I do.
________________________________

PR #303, opened 2020-03-04
Feature/pco

dinojugosloven:
Adding support for PCO camera


malcolm/modules/ADPco/blocks/pco_runnable_block.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+- ADPco.blocks.pco_driver_block:
+    mri: $(mri_prefix):DRV
+    prefix: $(pv_prefix):CAM
+
+- ADCore.parts.DetectorDriverPart:
+    name: DRV
+    mri: $(mri_prefix):DRV
+    soft_trigger_modes:
+        - Off
```
thomascobb: Is the trigger mode string for software triggering really "Off"?

PR #306, opened 2020-04-01
Improve PVA server logging

BenBradnick:
Just a small addition to re-add some logging for the PVA server comms when rpc() is called on BlockHandler.

What I wasn't sure about was whether I could get the logged parameters into a separate field on Graylog so it doesn't print the entire list in the main message.

I'm not sure if you want anything else logged here? This was sufficient to see the methods being called from GDA.

### DIFF
--- a/malcolm/core/process.py
+++ b/malcolm/core/process.py
@@ -216,7 +216,6 @@ class Process(Loggable):
         self._spawn_count = 0
         self._spawned = [s for s in self._spawned if not s.ready()]
 
-
     def add_controllers(self, controllers, timeout=None):
         # type: (List[Controller], float) -> None
         """Add many controllers to be hosted by this process
--- a/malcolm/imalcolm.py
+++ b/malcolm/imalcolm.py
@@ -202,7 +202,7 @@ def main():
 
     args = parse_args()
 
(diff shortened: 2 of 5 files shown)


thomascobb:
I think we should log once when an RPC is called (with its parameters) and once when it returns (with its return value)

BenBradnick:
I have finished making the changes. I noticed there isn't a test case explicitly for BlockHandler with unit tests for its methods (instead it looks like there is a system test in TestPVAServer). Would you like me to write one?

thomascobb:
Yes please, you only need to add test cases that improve coverage, if it's already covered by a system test then you don't need to add a unit test for it

BenBradnick:
The coverage for the BlockHandler shows the missing test areas are the nested functions. Do you know of an elegant solution to be able to test them?

I will also add tests for the PvaServerComms class

BenBradnick:
@thomascobb I believe I have finished writing the tests to improve coverage (after realising that patches apply in reverse order!). 

I couldn't think of a way to test the typing import, but otherwise all other paths now have tests.

thomascobb:
That was me, logged in as the wrong account!

PR #308, opened 2020-04-08
fix problem with sharksfin-like velocity time profile on high acceler…

gilesknap:
…ation motors

I'm starting the pull request now so you can take a look, but I will also add a test that proves the result tomorrow.

The issue was that the start point position was calculated in pmac_child_part.move_to_start using motor_info.ramp_distance without any minimum ramp time.

However when calculating the run up time later in pmac_child_part.calculate_generator_profile we were capping at a minimum .002s. Thus the 1st Prev->Current velocity calculation was out and threw the rest of them.

This is probably my record for the longest, most circuitous route to a simple solution (and I was already world champion)

### DIFF
--- a/malcolm/modules/pmac/infos.py
+++ b/malcolm/modules/pmac/infos.py
@@ -35,10 +35,12 @@ class MotorInfo(Info):
         ramp_time = abs(v2 - v1) / self.acceleration
         return ramp_time
 
-    def ramp_distance(self, v1, v2, ramp_time=None):
+    def ramp_distance(self, v1, v2, ramp_time=None, min_ramp_time=None):
         # The distance moved in the first part of the ramp
         if ramp_time is None:
             ramp_time = self.acceleration_time(v1, v2)
+        if min_ramp_time is not None:
+            ramp_time = max(ramp_time, min_ramp_time)
         ramp_distance = (v1 + v2) * ramp_time / 2
         return ramp_distance
 
--- a/malcolm/modules/pmac/parts/pmacchildpart.py
+++ b/malcolm/modules/pmac/parts/pmacchildpart.py
@@ -483,7 +483,9 @@ class PmacChildPart(builtin.parts.ChildPart):
             for _ in range(nsplit):
                 self.profile["timeArray"].append(time_point / nsplit)
             for _ in range(nsplit - 1):
-                self.profile["velocityMode"].append(VelocityModes.AVERAGE_PREV_TO_NEXT)
+                self.profile["velocityMode"].append(
+                    VelocityModes.AVERAGE_PREV_TO_NEXT
+                )
                 self.profile["userPrograms"].append(UserPrograms.NO_PROGRAM)
             for k, v in axis_points.items():
                 cs_axis = self.axis_mapping[k].cs_axis.lower()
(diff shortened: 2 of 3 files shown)


thomascobb:
This makes sense for a ramp at the start, but why are turnarounds also provoking the shark's fin case?

gilesknap:
Hmmm,

It did fix all the sharks fin effects and the reason is that all points are velocity mode 1 Prev->Current (except for the very first and last).

Thus any inaccuracies are propagated through the whole scan.

This is bad because the math needs to be dead accurate to work through and entire long scan. I'm guessing that this is unintentional but cant recall what is intended.

gilesknap:
Reviewing the unit tests today, I note that the standard configure test for pmac_child_part yields:
```velocityMode= [1, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 1, 3])```
i.e. prev->next in the row and prev->current in the turnarounds

But once we add `MotionTriggerInfo(MotionTrigger.NONE)]`. The rows loose all intermediate points and it becomes:
```velocityMode=pytest.approx([1, 1, 1, 1, 1, 1, 1, 3])```

So this is OK except that any problem in calculation will propagate the length of the scan. I also feel that any floating point rounding issues might accumulate over a very long scan.

Solution:
For the linear case it would be OK to use current->next for the starting point and it is only the linear case for which sparse trajectories apply.

**Comment by GDYendell** [comment]
Out of interest, why do you not use Current -> Next for row entry?

**Comment by gilesknap** [comment]
@GDYendell 
> Out of interest, why do you not use `Current -> Next` for row entry?

When the pmac is not triggering we use sparse trajectories to allow faster triggering than the pmac can do points at. So in a linear row we throw away all of the points except the first lower and and the last upper. Those always use `prev->current`, so we have ended up throwing away all `current->next` points.

**Comment by GDYendell** [comment]
If you only have two points per row, shouldn't it do:

C->N .......... P->C
C->N .......... P->C

so that the trajectory through all the points is 'horizontal'.

If you do P->C on the row entry, won't the velocity be this way through the first lower?:

                      o
               o
        o
L

**Comment by gilesknap** [comment]
Sorry I don't understand your ascii art but I think the confusion is over the fact that we have changed how P->C works and you may not know about it yet.

see https://confluence.diamond.ac.uk/x/Eh_vBQ in particular the heading Fix Previous to Next Calculation which shows the new math.

(@thomascobb This confirms the need to change the name of this velocity mode!)

**Comment by gilesknap** [comment]
@thomascobb and @gilesknap have agreed that the least invasive solution to this is to use the old averaging P->C velocity mode for the final point in a row.
There are only 2 bits used to encode the velocity mode, but luckily the old CURRENT->NEXT is no longer used. 

The naming has become confusing so I propose the following new names for the 4 modes:
```
class VelocityModes:
    AVERAGE_PREV_TO_NEXT = 0
    REAL_PREV_TO_CURRENT = 1
    AVERAGE_PREV_TO_CURRENT = 2
    ZERO_VELOCITY = 3```

gilesknap:
Just because the unit tests pass ...

The above checkin broke things:
image

gilesknap:
No wait. That is because I patched CURRENT->NEXT back in on my test rig. Nobody else will have that.

gilesknap:
Something is still up.

In latest released version of Prog 1 now running on my brick we have:
```
    ; Select velocity calculation based on VelMode value
    If(VelMode = 1)
        GoSub105                        ; Prev -> Current
    EndIf
    If(VelMode = 0)
        GoSub106                        ; Prev -> Next
    EndIf
    If(VelMode = 2)
        ; GoSub107                        ; Current -> Next
        ; TODO TODO - hack - this should be fixed Malcolm
        GoSub105
    EndIf
    If(VelMode = 3)
        GoSub108			; Zero
    EndIf
```
implying velmode 2 and velmode 1 do the same thing.

With the latest pymalcolm which puts velmode 2 at the end of a sparse row we see:
X/Y
image
t/v
image



But with the previous pymalocm which uses velmode 1 throughout:
X/Y
image
t/v
image

gilesknap:
That is a visualisation error. My plotting code still sees 2 as Current->Next. So all is well.

gilesknap:
I believe this is ready to merge.

My Malcolm system tests in pmac module pass with the new version of Prog1 and the old version of Prog1.

PR #310, opened 2020-06-19
Fix/hdfwriterpart on windows

BenBradnick:
This fixes the problem for Windows IOCs where Malcolm still creates the XML layout file with colons in the filename even though <HDF_PLUGIN>:XMLFileName will be appropriately converted to use underscores (in addition to the Windows drive letter).

- I have renamed the value of the layout file that is put to the PV over Channel Access
- The XML layout file is now created after the run on Windows check
- I have updated the MRI names used in the HDFWriterPart tests to use colons as would be typical for a real Beamline, and added a check in the configure test to ensure the colon is converted for Windows IOCs

### DIFF
--- a/malcolm/modules/ADCore/parts/hdfwriterpart.py
+++ b/malcolm/modules/ADCore/parts/hdfwriterpart.py
@@ -11,7 +11,7 @@ from malcolm.core import APartName, Future, Info, Block, PartRegistrar, \
 from malcolm.modules import builtin, scanning
 from ..infos import CalculatedNDAttributeDatasetInfo, NDArrayDatasetInfo, \
     NDAttributeDatasetInfo, FilePathTranslatorInfo
-from ..util import APartRunsOnWindows, FRAME_TIMEOUT, XmlLayout
+from ..util import APartRunsOnWindows, FRAME_TIMEOUT, make_xml_filename
 
 if TYPE_CHECKING:
     from typing import Iterator, List, Dict
@@ -344,7 +344,7 @@ class HDFWriterPart(builtin.parts.ChildPart):
         futures += set_dimensions(child, generator)
         xml = make_layout_xml(generator, part_info,
                               self.write_all_nd_attributes.value)
-        self.layout_filename = XmlLayout.get_layout_filename(
+        self.layout_filename = make_xml_filename(
             file_dir, self.mri, suffix="layout")
         with open(self.layout_filename, "w") as f:
             f.write(xml)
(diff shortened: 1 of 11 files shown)


thomascobb:
@dinojugosloven did something similar in #304 but I never checked it before it got merged. Can you check through that and see if you can work out what is best approach please?

BenBradnick:
So this uses the work from that branch by @dinojugosloven but this corrects an issue where Malcolm would still create the XML layout file with colons in the name based on the MRI before this path was translated to be Windows-friendly.

malcolm/modules/ADCore/parts/hdfwriterpart.py
```diff
@@ -346,14 +346,17 @@ def on_configure(self,
                               self.write_all_nd_attributes.value)
         self.layout_filename = os.path.join(
             file_dir, "%s-layout.xml" % self.mri)
-        with open(self.layout_filename, "w") as f:
-            f.write(xml)
-        layout_filename = self.layout_filename
+        layout_filename_pv = self.layout_filename
```
thomascobb, line 349: I would just do:
```python
self.layout_filename = os.path.join(
            file_dir, "%s-layout.xml" % self.mri.replace(":", "_")
```
And leave the rest of the code unchanged. I would also change the replaces in the FilePathTranslator into asserts, and put replaces in the stats plugin too.

thomascobb: This changes it to replace colons for underscores in Linux, but that's no bad thing...

BenBradnick:
@thomascobb Now colons are replaced by underscores independent of the platform. The translator part now only checks with an assert that there are no colons present in the filename.

As the XML layout filename code was essentially duplicated in 3 places (detector driver part, status plugin part and HDF writer part) I have moved this into a static method in a class in util.py.

If you're happy with these changes then this should be ready to be merged.

malcolm/modules/ADCore/util.py
```diff
@@ -92,3 +94,11 @@ def __init__(self,
         self.sourceType = ASourceTypes(sourceType)
         self.dataType = ADataTypes(dataType)
         self.datasetType = AAttributeTypes(datasetType)
+
+
+class XmlLayout:
+
+    @staticmethod
+    def get_layout_filename(file_dir, mri, suffix="attributes"):
```
thomascobb, line 102:
```suggestion
def make_xml_filename(file_dir, mri, suffix="attributes"):
    """Return a Block-specific filename for attribute or layout XML file"""
```
I'd dispense with the class, and just make this a plain old function

PR #314, opened 2020-07-20
Feature/xspress3v3

blondejamtart:
Update blocks for new xspress EPICS support (version 3). Moved old blocks to xspress3_DEPRECATED


Pipfile
```diff
@@ -33,6 +33,7 @@ numpy = "*"
 ruamel-yaml = "*"
 h5py = "==2.9.0"
 p4p = "*"
+packaging = "*"
```
thomascobb, line 36: Can we use setuptools.VersionInfo instead please?

blondejamtart: been looking at this...couldn't find VersionInfo in setuptools but did find some of the other versioning stuff within setuptools.
Currently we are checking that for the two versions (required and present), the major matches and the minor of the present is greater than or equal to that required. With the versioning stuff I found in setuptools (pkg_resources.parse_version() or setuptools.version.pkg_resources.packaging.version.Version) I can't find a way to do this sort of comparison.

thomascobb: ok, lets leave it like it is then

PR #315, opened 2020-10-14
make ADTetrAMM consider point duration and a exposure dead time

hir12111:
This change resolves [jira issue I22-1027](https://jira.diamond.ac.uk/browse/I22-1027).
As tetrAMM AD driver is not a standard one (it misses a few expected PVs), the following tricks were used:
- exposure attribute is using the averaging time PV, which is a good approximation, as it is the time in which the device will be taking samples
- The acquire period was added as a Float64Part in order to bypass the error in which the parent class (DetectorDriverPart) tries to set it.
Another approach could be modifying the DetectorDriverPart class, but because the tratrAMM is probably the only exception needing it, I thought it was better to have a 'dummy' part and keep the code as it is.

I've marked this PR as WIP because it hasn't been tested for python3 Pymalcolm, but it will be done next shutdown.

### DIFF
--- a/malcolm/modules/ADTetrAMM/blocks/tetrAMM_driver_block.yaml
+++ b/malcolm/modules/ADTetrAMM/blocks/tetrAMM_driver_block.yaml
@@ -35,3 +35,21 @@
     pv: $(pv_prefix):TriggerMode
     rbv_suffix: _RBV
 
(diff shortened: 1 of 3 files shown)


hir12111:
This was already tested in I22 using python3, a new attribute was added in order to have some control over the number of samples in each frame

PR #317, opened 2020-11-06
Feature/directory monitor

BenBradnick:
This adds a directory monitor block and part to check whether all of the expected file mounts and file mount types are correct for a particular server. The part checks during configure and by default will go into error (but otherwise can be set to just log an error in case you want the scan to proceed anyway with potential performance issues).

Also fixed the pandasettrigger for new version of isort and tidied up the AttributePreRunPart test file.

### DIFF
--- a/malcolm/modules/scanning/parts/directorymonitorpart.py
+++ b/malcolm/modules/scanning/parts/directorymonitorpart.py
@@ -1,0 +1,42 @@
+from annotypes import Anno, add_call_types
+
+from malcolm.core import PartRegistrar
+from malcolm.modules.builtin.hooks import AContext
+from malcolm.modules.builtin.parts import AMri, APartName, ChildPart
+
+from .. import hooks
+
+with Anno("Whether to raise a ValueError for a bad status"):
+    AErrorOnFail = bool
+
+
+class DirectoryMonitorPart(ChildPart):
+    """Part for checking a directoryMonitor Manager is happy"""
+
+    def __init__(
+        self, name: APartName, mri: AMri, error_on_fail: AErrorOnFail = True
+    ) -> None:
+        super().__init__(name, mri, initial_visibility=True)
+        self.error_on_fail = error_on_fail
+
+    def setup(self, registrar: PartRegistrar) -> None:
+        super().setup(registrar)
+        # Hooks
+        registrar.hook(hooks.ConfigureHook, self.check_directories)
+
+    @add_call_types
+    def check_directories(self, context: AContext) -> None:
+        child = context.block_view(self.mri)
+        # Perform check
+        getattr(child, "managerCheck")()
+
+        # Get status
+        status = getattr(child, "managerStatus").value
+        if not status:
+            hostname = getattr(child, "managerHostname").value
+            bad_status_string = (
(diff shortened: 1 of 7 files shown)


malcolm/modules/scanning/parts/directorymonitorpart.py
```diff
@@ -0,0 +1,42 @@
...
+    def setup(self, registrar: PartRegistrar) -> None:
+        super().setup(registrar)
+        # Hooks
+        registrar.hook(hooks.ConfigureHook, self.check_directories)
+
+    @add_call_types
+    def check_directories(self, context: AContext) -> None:
+        child = context.block_view(self.mri)
+        # Perform check
+        getattr(child, "managerCheck")()
```
thomascobb, line 31:
```suggestion
        child.managerCheck()
```

malcolm/modules/scanning/parts/directorymonitorpart.py
```diff
@@ -0,0 +1,42 @@
...
+
+        # Get status
+        status = getattr(child, "managerStatus").value
```
thomascobb, line 34:
```suggestion
        status = child.managerStatus.value
```
etc.

malcolm/modules/scanning/parts/directorymonitorpart.py
thomascobb: Does this support caput callback? There might still be a race condition here between getting the callback and reading the monitored managerStatus value. You might want to put in a 0.1s sleep (ugg) to guard against this

thomascobb (changes requested):

malcolm/modules/scanning/parts/directorymonitorpart.py
BenBradnick: It does, so the status value should be updated before it gets checked (assuming that just calling managerCheck() will wait for this callback, or does Malcolm not wait for the callback by default?).

BenBradnick: Looks like it does wait by default in the CAActionPart, and I didn't encounter any problems in testing (I changed the file mounts to a bad/good state and these changes were picked up by this part on configure - without any additional polling in place)

thomascobb: Malcolm monitors all PVs. There is a race condition between the caput callback returning and the monitored PV updating. This is a CA limitation. You can only get around this by cagetting the PV rather than monitoring it. 

Actually, this is exactly what the CAActionPart does. If you set the `status_pv` of managerCheck to whatever managerStatus points to, then it will do this check for you and raise an AssertionError if `status_pv` doesn't match

malcolm/modules/scanning/parts/directorymonitorpart.py
BenBradnick: Not even sure what I was thinking here...

malcolm/modules/scanning/parts/directorymonitorpart.py
BenBradnick: I see, the monitoring is the problem. I have now refactored to use the status_pv and checking it against the good value.

thomascobb (approved):

PR #319, opened 2020-11-18
Removed syslog handler to prevent excessive logging to local disk

EdWarrick:
Fix for an issue on I14 - log writing to syslog was filling up the local disk. We don't currently use syslog logs for anything so this simply disables this logging handler.


thomascobb:
Can you resolve the conflicts please?

PR #320, opened 2020-12-18
Feature/adethercat

BenBradnick:
This PR adds the following:

ADEthercat:

- Driver block and runnable block with a standard set of plugins and a reframe_plugin_block
- Driver part which is based on DetectorDriverPart, doesn't have any progress and runs continuously using a software trigger mode. Is armed during PreRunHook

ADCore:

- reframe_plugin_block: for the Reframe plugin parts
- ReframePluginPart: works a bit like a DetectorDriverPart, but simpler. Inherits from ChildPart, slightly different PVs used to control acquisition (because it is a AreaDetector plugin and not a "driver")
- HDFWriterPartWithTimeout: Inherits from HDFWriterPart, but instead of having a finite end it, the part continues until it detects no new frames within a timeout provided in the constructor.

The intention is that the HDFWriterPartWithTimeout sits outside the block(s) controlling the detector/upstream plugin producing the frames, which then stops after the set number of images is taken. This allows the HDF5 plugin to write any number of frames without having to equal to the number of steps in the scan.

One thing I was unsure of was whether there should be a timeout for waiting for the HDF5 plugin queue to reach 0 after on_run() has determined the plugin has finished writing frames. Also whether the part should be in ADCore or ADEthercat, as it probably only has a use case for Ethercat currently.


thomascobb (commented):
Discussed on zoom

BenBradnick:
The original PR has been refactored to stick to standard parts/blocks as much as possible. I had to make some changes to the existing ReframePluginPart and also move where the DetectorDriverPart determines if the detector is in a software triggered state to above setup_detector() (as the ReframePluginPart uses the attribute in setup_detector()).

The ReframePluginPart uses the trigger on condition to determine whether the trigger "mode" is software or hardware. This along with the trigger off condition are saved in the design so it should be possible to swap between software/hardware modes by loading an appropriate design.

Now the continuous block is expected to be in a completely separate scan block which is run independently of the runnable block. This is okay as the only shared component is the EthercatDriverPart, which only starts the acquisition and doesn't stop on abort/reset. The generator duration controls the size of the frame passed on from the reframe plugin based on the sample frequency set on the part - so e.g. 0.001s duration for a 1kHz sample rate results in 1 sample per array, 1s duration results in 1000 samples per array.

The two ethercat runnable blocks are now almost identical except for an ExposureDeadtimePart in the hardware block - this is required for the hardware block to avoid missing triggers, but adding it to the continuous block prevents doing 1 sample per frame. Whilst I can override the readout time and frequency accuracy of the part to get it working for the continuous block, this seemed a bit messy.

thomascobb (approved):

PR #321, opened 2021-01-05
Add support for "3D Scans"

GDYendell:
This is an initial partially working version. I thought this would be a good point to ask if what I have done so far is reasonable. I can run 3x100 and Nx1 scans manually through the web GUI with no problems. But when we tested it with GDA, it got very upset (41 point step scan, 0.5 second exposure, ~1 second per run):

```
removed
```

This, or something similar, happens everytime, but half the time the scan does actually finish as expected. The terminal continues partially logging if you hit enter, but the prompt is gone and if I try to run any more scans it doesn't do anything.

Have I done something insane? I guess the most obvious thing is the try catch I have added in the timeout loop. Or, maybe it is unrelated.


GDYendell:
It looks like there is some overlap between this and #290. Do we need both?

malcolm/modules/ADCore/parts/detectordriverpart.py
```diff
@@ -92,6 +92,7 @@ def setup_detector(
         context: Context,
         completed_steps: scanning.hooks.ACompletedSteps,
         steps_to_do: scanning.hooks.AStepsToDo,
+        steps_in_run: int,
```
thomascobb, line 95: scanning.hooks.AStepsInRun

GDYendell:
I have take out my hacky try except timeout and used event_timeout instead and the scan works without the horrible messages and breaking the prompt, through GDA.

There is still an issue with the completion status, which I think confuses GDA:

image
image

GDYendell:
Web GUI fixed. Apparently GDA might be confused with progress for other reasons.

GDYendell:
I have updated all the tests except the one that is failing. This seems like an unrelated issue. Does anyone know what the problem might be here?

```
__________________ TestXmap3DetectorDriverPart.test_configure __________________

Traceback (most recent call last):
  File "/home/travis/build/dls-controls/pymalcolm/tests/test_modules/test_xmap/test_xmapdriverpart.py", line 74, in test_configure
    assert self.child.handled_requests.mock_calls == [
AssertionError: assert [call.put('arrayCallbacks', True),\n call.put('arrayCounter', 0),\n call.put('autoPixelsPerBuffer', 'Manual'),\n call.put('binsInSpectrum', 2048),\n call.put('collectMode', 'MCA mapping'),\n call.put('ignoreGate', 'No'),\n call.put('inputLogicPolarity', 'Normal'),\n call.put('pixelAdvanceMode', 'Gate'),\n call.put('pixelsPerBuffer', 1),\n call.put('pixelsPerRun', 456),\n call.put('presetMode', 'No preset'),\n call.put('attributesFile', 'Z:\\mri-attributes.xml'),\n call.post('start'),\n call.when_value_matches('acquiring', True, None)] == [call.put('arrayCallbacks', True),\n call.put('arrayCounter', 0),\n call.put('autoPixelsPerBuffer', 'Manual'),\n call.put('binsInSpectrum', 2048),\n call.put('collectMode', 'MCA mapping'),\n call.put('ignoreGate', 'No'),\n call.put('inputLogicPolarity', 'Normal'),\n call.put('pixelAdvanceMode', 'Gate'),\n call.put('pixelsPerBuffer', 1),\n call.put('pixelsPerRun', 456),\n call.put('presetMode', 'No preset'),\n call.post('start'),\n call.when_value_matches('acquiring', True, None),\n call.put('attributesFile', 'Z:\\mri-attributes.xml')]

  At index 11 diff: call.put('attributesFile', 'Z:\\mri-attributes.xml') != call.post('start')

  Full diff:
    [
     call.put('arrayCallbacks', True),
     call.put('arrayCounter', 0),
     call.put('autoPixelsPerBuffer', 'Manual'),
     call.put('binsInSpectrum', 2048),
     call.put('collectMode', 'MCA mapping'),
     call.put('ignoreGate', 'No'),
     call.put('inputLogicPolarity', 'Normal'),
     call.put('pixelAdvanceMode', 'Gate'),
     call.put('pixelsPerBuffer', 1),
     call.put('pixelsPerRun', 456),
     call.put('presetMode', 'No preset'),
  +  call.put('attributesFile', 'Z:\\mri-attributes.xml'),
     call.post('start'),
     call.when_value_matches('acquiring', True, None),
  -  call.put('attributesFile', 'Z:\\mri-attributes.xml'),
    ]
```

Next I will add some new tests.

GDYendell:
I added a commit to fix mypy, but I noticed I didn't have to do the same elsewhere. It is because in `PositionLabellerPart` `done_when_reaches` is defined as None, but in `DetectorDriverPart` it is defined as `0`. Should this be made consistent @thomascobb ?

BenBradnick:
> 
> 
> I have updated all the tests except the one that is failing. This seems like an unrelated issue. Does anyone know what the problem might be here?
> 
> ```
> __________________ TestXmap3DetectorDriverPart.test_configure __________________
> 
> Traceback (most recent call last):
>   File "/home/travis/build/dls-controls/pymalcolm/tests/test_modules/test_xmap/test_xmapdriverpart.py", line 74, in test_configure
>     assert self.child.handled_requests.mock_calls == [
> AssertionError: assert [call.put('arrayCallbacks', True),\n call.put('arrayCounter', 0),\n call.put('autoPixelsPerBuffer', 'Manual'),\n call.put('binsInSpectrum', 2048),\n call.put('collectMode', 'MCA mapping'),\n call.put('ignoreGate', 'No'),\n call.put('inputLogicPolarity', 'Normal'),\n call.put('pixelAdvanceMode', 'Gate'),\n call.put('pixelsPerBuffer', 1),\n call.put('pixelsPerRun', 456),\n call.put('presetMode', 'No preset'),\n call.put('attributesFile', 'Z:\\mri-attributes.xml'),\n call.post('start'),\n call.when_value_matches('acquiring', True, None)] == [call.put('arrayCallbacks', True),\n call.put('arrayCounter', 0),\n call.put('autoPixelsPerBuffer', 'Manual'),\n call.put('binsInSpectrum', 2048),\n call.put('collectMode', 'MCA mapping'),\n call.put('ignoreGate', 'No'),\n call.put('inputLogicPolarity', 'Normal'),\n call.put('pixelAdvanceMode', 'Gate'),\n call.put('pixelsPerBuffer', 1),\n call.put('pixelsPerRun', 456),\n call.put('presetMode', 'No preset'),\n call.post('start'),\n call.when_value_matches('acquiring', True, None),\n call.put('attributesFile', 'Z:\\mri-attributes.xml')]
> 
>   At index 11 diff: call.put('attributesFile', 'Z:\\mri-attributes.xml') != call.post('start')
> 
>   Full diff:
>     [
>      call.put('arrayCallbacks', True),
>      call.put('arrayCounter', 0),
>      call.put('autoPixelsPerBuffer', 'Manual'),
>      call.put('binsInSpectrum', 2048),
>      call.put('collectMode', 'MCA mapping'),
>      call.put('ignoreGate', 'No'),
>      call.put('inputLogicPolarity', 'Normal'),
>      call.put('pixelAdvanceMode', 'Gate'),
>      call.put('pixelsPerBuffer', 1),
>      call.put('pixelsPerRun', 456),
>      call.put('presetMode', 'No preset'),
>   +  call.put('attributesFile', 'Z:\\mri-attributes.xml'),
>      call.post('start'),
>      call.when_value_matches('acquiring', True, None),
>   -  call.put('attributesFile', 'Z:\\mri-attributes.xml'),
>     ]
> ```
> 
> Next I will add some new tests.

I think this is caused by the move of arm_detector() in the DetectorDriverPart's on_configure() method, so that now it is after setting the attributes up - so the order of calls is now slightly different is all.

thomascobb:
@GDYendell mentioned this in #321

GDYendell:
> the move of arm_detector() in the DetectorDriverPart's on_configure()
Of course. Thanks @BenBradnick

GDYendell:
I added a commit to fix mypy, but I noticed I didn't have to do the same elsewhere. It is because in `PositionLabellerPart` `done_when_reaches` is defined as None, but in `DetectorDriverPart` it is defined as `0`. Should this be made consistent?

thomascobb:
> I added a commit to fix mypy, but I noticed I didn't have to do the same elsewhere. It is because in `PositionLabellerPart` `done_when_reaches` is defined as None, but in `DetectorDriverPart` it is defined as `0`. Should this be made consistent?

Yes, we should make it consistent. In this case they can both be defined as 0, as we always set it in configure.

malcolm/modules/ADCore/parts/positionlabellerpart.py
```diff
@@ -83,7 +83,7 @@ def on_configure(
         futures += child.put_attribute_values_async(
             dict(enableCallbacks=True, idStart=id_start)
         )
-        self.steps_up_to = completed_steps + steps_to_do
+        self.steps_up_to = generator.size
```
GDYendell, line 82: I am not sure if this is correct when completed steps is not 0. Initially I changed it to
```
self.steps_up_to = completed_steps + generator.size
```
but then I got errors trying to access generator points with index > generator.size, so I changed it.

GDYendell:
I have added some tests for the functionality this adds, so I think this is ready to look at again @thomascobb .
Initially I thought they would be tests where I call configure, run, run etc and check behaviour based on what the frame counter does, but it does not look like this is done in the existing tests. I don't know if there are any examples of system-ish tests like that I could look at to add something similar here.

I also have some local black failures for lines I didn't change (imalcolm.py and tetrammdriverpart.py), but the travis build seems happy, so I am not sure whether to add them here.

gilesknap:
re black: I noticed that recent versions have changed some subtle rules (I've already forgotten what) this may explain your experience?

BenBradnick:
> 
> 
> re black: I noticed that recent versions have changed some subtle rules (I've already forgotten what) this may explain your experience?

Yeah the current builds of Travis are using black 20.x, which has some breaking changes for formatting rules - I had the same issue

malcolm/modules/ADCore/parts/detectordriverpart.py
```diff
@@ -144,13 +145,19 @@ def wait_for_detector(
         child.arrayCounterReadback.subscribe_value(
             self.update_completed_steps, registrar
         )
-        # If no new frames produced in event_timeout seconds, consider scan dead
-        context.wait_all_futures(self.start_future, event_timeout=event_timeout)
+
+        if not self.is_hardware_triggered:
+            # If soft triggered then wait for the detector to stop acquiring before
+            # checking if all frames were received
+            context.wait_all_futures(self.start_future, event_timeout=event_timeout)
```
thomascobb, line 152: I reckon we should never bother storing and checking this start_future, we can't check it in the hardware triggered case so there's not much to be gained.

malcolm/modules/ADCore/parts/detectordriverpart.py
```diff
@@ -144,13 +145,19 @@ def wait_for_detector(
...
+
         # Now wait to make sure any update_completed_steps come in. Give
         # it 5 seconds to timeout just in case there are any stray frames that
         # haven't made it through yet
         child.when_value_matches(
-            "arrayCounterReadback", self.done_when_reaches, timeout=DEFAULT_TIMEOUT
+            "arrayCounterReadback",
+            self.done_when_reaches,
+            event_timeout=DEFAULT_TIMEOUT,
```
thomascobb, line 160:
```suggestion
            event_timeout=event_timeout,
```

thomascobb: In case we have a slow scan

thomascobb (changes requested):

malcolm/modules/ADCore/parts/detectordriverpart.py
GDYendell: Removed in 1d1488e

GDYendell:
I think this is ready to look at again @thomascobb

malcolm/modules/ADCore/parts/detectordriverpart.py
```diff
@@ -150,13 +151,14 @@ def wait_for_detector(
         child.arrayCounterReadback.subscribe_value(
             self.update_completed_steps, registrar
         )
-        # If no new frames produced in event_timeout seconds, consider scan dead
-        context.wait_all_futures(self.start_future, event_timeout=event_timeout)
+
         # Now wait to make sure any update_completed_steps come in. Give
         # it 5 seconds to timeout just in case there are any stray frames that
         # haven't made it through yet
```
thomascobb, line 157:
```suggestion
        # Wait for the array counter to reach the desired value. If any one
        # frame takes more than event_timeout to appear, consider scan
        # dead
```

malcolm/modules/ADCore/parts/detectordriverpart.py
```diff
@@ -220,15 +222,12 @@ def setup(self, registrar: PartRegistrar) -> None:
         # Hooks
         registrar.hook(scanning.hooks.ReportStatusHook, self.on_report_status)
         registrar.hook(
-            (
-                scanning.hooks.ConfigureHook,
-                scanning.hooks.PostRunArmedHook,
-                scanning.hooks.SeekHook,
-            ),
+            (scanning.hooks.ConfigureHook, scanning.hooks.SeekHook,),
```
thomascobb, line 225:
```suggestion
            (scanning.hooks.ConfigureHook, scanning.hooks.SeekHook),
```

tests/test_modules/test_ADCore/test_detectordriverpart.py
```diff
@@ -154,6 +154,32 @@ def test_run(self):
         assert self.o.registrar.report.call_count == 2
         assert self.o.registrar.report.call_args[0][0].steps == 0
 
+    def test_hardware_triggered_run(self):
+        self.o.registrar = MagicMock()
```
thomascobb, line 158: What is this test testing for?

malcolm/modules/ADCore/parts/hdfwriterpart.py
```diff
@@ -557,6 +556,12 @@ def on_run(self, context: scanning.hooks.AContext) -> None:
             else:
                 return
 
+    @add_call_types
+    def on_post_run_armed(
+        self, context: scanning.hooks.AContext, steps_to_do: scanning.hooks.AStepsToDo,
+    ) -> None:
+        self.done_when_reaches += steps_to_do
```
thomascobb, line 563: Why do we need this rather than sticking with calling on_seek on post run armed?

tests/test_modules/test_ADCore/test_detectordriverpart.py
GDYendell: It tests that we do not call arm_detector in run() if we are hardware triggered

malcolm/modules/ADCore/parts/hdfwriterpart.py
GDYendell: on_seek fiddles with the UID and resets the arrayCounter. We don't want that. When we get into PostRunArmed with 100/300 done, we just want to say, OK now you need to get to 200/300, run again.

malcolm/modules/ADCore/parts/positionlabellerpart.py
GDYendell: Any thoughts on this @thomascobb ?

malcolm/modules/ADCore/parts/detectordriverpart.py
```diff
@@ -299,29 +297,38 @@ def on_configure(
...
             self.is_hardware_triggered = mode not in self.soft_trigger_modes
+
+        # If detector is hardware triggered we can configure the detector for all frames
+        # now, rather than configuring and arming for each inner scan, and send the
+        # triggers for one inner scan in each run
+        steps_in_run = steps_to_do
         if self.is_hardware_triggered:
-            # Start now if we are hardware triggered
-            self.arm_detector(context)
+            steps_to_do = generator.size
```
thomascobb, line 320:
```suggestion
            steps_to_do = generator.size - completed_steps
```

malcolm/modules/ADCore/parts/positionlabellerpart.py
thomascobb: ```self.done_when_reaches = genertor.size - completed_steps```

malcolm/modules/ADCore/parts/detectordriverpart.py
thomascobb: we decided to call it ```steps_until_done```

GDYendell: Instead of renaming this I have decided to take out the new steps_in_run parameter entirely and maintain this in steps_to_do, because that is what it represents in on_configure. What I really wanted here is to be able to set numImages to a value other than steps_to_do, i.e. the whole scan rather than the inner scan, so I have just added a num_images parameter instead. I think this is much clearer. We could consider changing steps_to_do to steps_in_run, though.

GDYendell:
Rebased again to resolve conflicts. @thomascobb I think this is ready to look at again. Note the comment above in reference to renaming steps_until_done, which is now redundant.

thomascobb (approved):

prefix/bin/imalcolm
```diff
@@ -0,0 +1,11 @@
+#!/dls_sw/work/mef65357/python3/RHEL7-x86_64/pymalcolm/lightweight-venv/bin/python
```
BenBradnick, line 1: I presume this file is not supposed to be committed?

PR #324, opened 2021-01-28
Feature/GitHub actions

BenBradnick:
Add a workflow for GitHub Actions

To do:

- Investigate caching for EPICS base (currently it downloads and runs make every time, takes ~60s)
- Investigate caching for the pipenv - a bit more difficult. Online examples show using the lock file, but this is currently skipped because some modules only contain hashes from what appear to be the Diamond PyPI repo. What is the correct approach for this?
- Check if deployment to pypi should be added
- Check if we need core dumps


BenBradnick:
EPICS Base should now be being cached and restored based on a key tied to the version installed.

Currently it takes ~1 minute to install the dependencies in the Pipfile and 3-4 minutes to run the tests and linter. The biggest difference so far is the response time is far better than Travis (the queue time is typically < 1 minute).

Currently it only builds for 3.7. Would there be a benefit adding other versions?

PR #330, opened 2021-02-11
Fix/imalcolm and docs

### LINKED ISSUE #322: CONTRIBUTING.rst out of date
Now that we have moved to using pipenv by default I presume this should also be updated to reflect this change.

- Update running the tests to use Pipfile and pipenv
- Code styling to mention black, flake8 and isort (also mention mypy?)
- Update how to build docs

One thing I noticed was that it mentioned using dashes for release numbers until Python 3. Should we now swap to using periods?

BenBradnick:
This will hopefully bring the documentation up to date for installing, developing and using Malcolm now that we have moved to pipenv.

This should fix #322.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,23 @@ Change Log
 All notable changes to this project will be documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`_ after 2-1.
 
+`Unreleased`_ - 2021-02-11
+--------------------------
+
+Added:
+
+- ADEthercat support - two runnable blocks for hardware and software-triggered
+  devices
+- Breakpoints - these can be defined for the scan to be broken up into
+  multiple sections where run returns in an armed state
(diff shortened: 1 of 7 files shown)


thomascobb (approved):

BenBradnick:
Just one comment:

Martin noticed the following TODO left in CONTRIBUTING.RST:
```
TODO: add convenience for supplying private properties as per MotionChildPart
```
What should this be replaced with?

thomascobb:
@gilesknap your name is on the commit that added that line, can you remember what it was about?

gilesknap:
Sorry, I have no idea what that was, luckily it as the word convenience in it so we can remove it without forgetting anything essential I guess!
________________________________

BenBradnick:
I think this is pretty close to the Python 3 template module now.

At some point the documentation probably wants to be restructured based on the template, but I will do that at a later date.

thomascobb:
looks good

PR #331, opened 2021-02-18
fix (?) kinematics parts to write all Q vars to VDS 

blondejamtart:
Comment seems to suggest it was meant to, nice to have for off-axis error checking.
Alternative or additional change would be to have all known scannables which are being transformed linked and named in the VDS (currently only the active axes are).
Almost definitely won't get time to merge this before leaving, so will likely need to be picked up.


thomascobb:
It looks harmless enough. @BenBradnick do you have anything using kinematics at the moment to test this with?

BenBradnick:
@thomascobb I don't have anything currently, I think this will be waiting on Huw to be able to test on J13.

BenBradnick:
I should now be able to test this at some point as part of the new I18 sample stage work.

PR #333, opened 2021-03-09
Update tutorials to use pipenv with Python 3

BenBradnick:
Updated the tutorials based on using pipenv.

I have also updated the pmac's shebang to use the default Malcolm version configured from configure-tool, as well as comment on the alternatives available.

### DIFF
--- a/docs/tutorials/pmac.rst
+++ b/docs/tutorials/pmac.rst
@@ -70,7 +70,7 @@ Define the Process Definition
 
 Let's make a Process Definition in ``etc/malcolm/BLxxI-ML-MALC-01.yaml``::
 
-    #!/dls_sw/prod/common/python/RHEL7-x86_64/pymalcolm/4-x/prefix/bin/imalcolm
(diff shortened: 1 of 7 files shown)


docs/tutorials/pmac.rst
```diff
@@ -70,7 +70,7 @@ Define the Process Definition
 
 Let's make a Process Definition in ``etc/malcolm/BLxxI-ML-MALC-01.yaml``::
 
-    #!/dls_sw/prod/common/python/RHEL7-x86_64/pymalcolm/4-x/prefix/bin/imalcolm
+    #!/dls_sw/prod/tools/RHEL7-x86_64/defaults/bin/imalcolm
```
thomascobb, line 73: I'd prefer to keep the version number in there in examples too, to stop people copy and pasting the wrong thing...

BenBradnick: I have now pointed it at release 4.4 in the example.

thomascobb (approved):

PR #336, opened 2021-03-23
Feature/beam selector different exposures

BenBradnick:
Work in progress for improving the beam selector scan.

- Convert attributes to using snake case
- Create a utility function for getting the minimum turnaround and interval information (as it is used in 3 separate places)
- Use information in detector table to determine delay_after
- Add mutator to modify delay_after for alternating points in the generator
- CompoundGenerator duration determined by move time and exposure times (or minimum turnaround in place of exposure times if larger)

I have also written some additional test cases.

Things to check:

- Is the pmac trajectory actually as expected? The unit tests "pass" but I need to verify this is the actual trajectory I want.


BenBradnick:
I have tested the branch on a beamline and believe this is ready for a review. Recent changes include:

- Making the PandA AlternatingDivPart for checking frames_per_step was 2 generic so that it can handle a single or list of valid frames_per_step.
- Now generator duration is calculated if 0 is passed. When validating the parameters, if the generator is tweaked this is resolved first before other tweaks. This helps ensure that if an exposure time has to be guessed, then that is tweaked after the final generator duration is known to behave similarly as before (in terms of using the longest exposure time for a given duration). This should work for guessing the duration for multiple detectors, and the largest tweak to the duration is applied.
- Added an option to set the delay to 0.0s for the PandA pulse trigger part (defaults to not being enabled). This is needed for the beam selector scan.

What should be checked:

1. Whether util.py in pmac is a good place for the mutator used by the beam selector scan
2. Whether the pmacChildPart should handle tweaking the generator duration based on the speed of the axes being used, rather than just the minimum time of 2ms.

malcolm/modules/scanning/parts/multipletriggerpart.py
```diff
@@ -0,0 +1,90 @@
...
+class MultipleTriggerPart(builtin.parts.ChildPart):
+    """Part for informing a parent that it can have multiple
+    triggers.
+    """
+
+    def __init__(
+        self,
+        name: APartName,
+        mri: AMri,
+        valid_multiples: Union[AMultiples, Sequence[int], int] = None,
```
thomascobb, line 26: When would we need anything other than 2 in this list?

BenBradnick: Good question! I thought potentially with 3 beam positions on K11 I'd keep it generic enough to be able to change it in the future.

In hindsight I can't think of what I would want a list for - as I'd have to load a different design if swapping between a different number of frames per step.

thomascobb: I think you'd only want a divide into 3 if you had 3 detectors, and I hope we never have a beamline with 3 beams...

BenBradnick: I guess K11 has "three" beams but 2 (pink and mono) should only be used one at a time.

Should I modify this to only handle a fixed value of 2 and rename the part to DoubleTriggerPart? DoubleFramesPerStepPart?

I named it a more generic term in the first place because of EtherCAT also wanting these doubling triggers.

thomascobb: That makes sense

BenBradnick: The part is now called DoubleTriggerPart and checks that frames_per_step is always equal to 2.

thomascobb:
> The main concern for me is if the correct design is not loaded on the brick so it cannot find the axes - should Validate still pass, and not tweak the duration, or throw an error complaining?

I think we have skipped some checks in validate because of this issue, so I'd probably let it pass

> The calculation of the generator duration tweak is also slightly different - I wanted it to always increase instead of decrease otherwise the validation of detectors would fail.

That sounds reasonable

PR #340, opened 2021-04-08
Fixes

thomascobb:
Two things
- update_part_exportable() was failing when response was a Return, probably harmless but good to fix it anyway
- better error message for when PandA disconnects

### DIFF
--- a/malcolm/modules/builtin/parts/childpart.py
+++ b/malcolm/modules/builtin/parts/childpart.py
@@ -257,7 +257,7 @@ class ChildPart(Part):
         child = self.child_controller.block_view()
         spawned = []
         if isinstance(response, Update):
-            new_fields = response.value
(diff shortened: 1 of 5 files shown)


malcolm/modules/builtin/parts/childpart.py
```diff
@@ -257,7 +257,7 @@ def update_part_exportable(self, response: Response) -> None:
         child = self.child_controller.block_view()
         spawned = []
         if isinstance(response, Update):
-            new_fields = response.value
+            new_fields = list(response.value)  # type: ignore
```
BenBradnick, line 260: Will value potentially be None here? If not, an assert could be used instead.

BenBradnick (approved):
Mypy now passes and I rebased on master to stop codecov complaining about a decrease in coverage.

PR #341, opened 2021-04-23
Restore default ChildPart.initial_visibility=None

thomascobb:
This means "True if the child Block is connected to anything"

### DIFF
--- a/tests/test_modules/test_builtin/test_childpart.py
+++ b/tests/test_modules/test_builtin/test_childpart.py
@@ -45,7 +45,6 @@ class TestChildPart(unittest.TestCase):
             mri=block_mri,
             name="part%s" % block_mri,
             stateful=False,
-            initial_visibility=True,
         )
         self.p.add_controller(controller)
         return part, controller
@@ -106,6 +105,7 @@ class TestChildPart(unittest.TestCase):
         assert info_out.name == "sourceportConnector"
         assert info_out.port == Port.INT32
         assert info_out.connected_value == "Connector1"
+        assert self.c.block_view().layout.value.visible == [True, True, True]
 
     def test_layout(self):
         b = self.p.block_view("mainBlock")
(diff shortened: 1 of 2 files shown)


BenBradnick (approved):
Looks good, sorry for messing this up trying to make mypy happy!

I see that the layout already has some more detailed tests with parts, so I've just added a couple of tests based on checking the visibility attribute is initialised in the first call to on_layout.

Do you thing any other testing is required? Happy to add more if so.

tests/test_modules/test_builtin/test_childpart.py
```diff
@@ -105,6 +105,7 @@ def test_init(self):
         assert info_out.name == "sourceportConnector"
         assert info_out.port == Port.INT32
         assert info_out.connected_value == "Connector1"
+        assert self.c.block_view().layout.value.visible == [True, True, True]
```
thomascobb, line 108: This test fails on master and passes here, so this should be sufficient (with the initial_visibility=True removed above)

tests/test_modules/test_builtin/test_childpart.py
```diff
@@ -154,3 +155,37 @@ def test_load_save(self):
         assert structure2 == expected
         self.p1.on_load(context, dict(sinkportConnector="blah_again"))
         assert b1.sinkportConnector.value == "blah_again"
+
+    def test_on_layout_first_call_sets_part_visibility_to_False(self):
```
thomascobb, line 159: I'm not a big fan of mocked tests, I try to only use them to test the last few corner cases, most of the time I try and cover with integration tests like the rest of this file. I'll leave it up to you whether these add anything or not...

BenBradnick: Sure, I agree that it makes a lot of sense to use the integration test which now covers this check anyway.

thomascobb:
Looks good

PR #344, opened 2021-06-02
Fix/hardware trigger breakpoints

BenBradnick:
The changes with arming a hardware-triggered detector for all points even with outer axes seems to break the breakpoints interface, at least for start of row triggering.

It also breaks software-triggered detectors as it makes the assumption that each breakpoint is after the same number of steps.

These changes revert the behaviour back to breaking down the scan into chunks when breakpoints are included, but should hopefully leave the behaviour untouched where breakpoints aren't provided.


thomascobb (approved):

GDYendell (approved):
This still works as I would expect on I07

PR #345, opened 2021-06-30
Fix/i18 deployment

BenBradnick:
Fixes for some issues discovered in I18 deployment testing:

- PositionLabellerPart not uploading additional points after initial 5,000, and now waits until unique ID reaches last ID expected using same logic as DetectorDriverPart. This was working using the AreaDetector simulator with pausing, breakpoints and pausing with breakpoints
- Savu kinematics can now scan without any axes in AxesToMove for no motion and added NeXus templates to the package data
- Cast data type in OdinWriterPart to lower case for NumPy V1.2 update

Then for convenience I added a basic block for the UnrollingPart so that it could be enabled/disabled in a design based on visibility. This allows the same scan block YAML to be used by Odin and non-Odin detectors.


BenBradnick:
I found additional issues when using a software-triggered detector and outer axes. If I paused part way through an inner scan, then the number of images to take would be updated for the remainder of that inner scan in Seek and then not be reset in PostRunArmed. This should now be fixed.

I want to do more testing with hardware triggering with scans with both breakpoints and outer axes with pausing to confirm these are all working together.

PR #348, opened 2021-08-10
Use servo frequency PV instead of I10

lyodaom:
This way, the PmacStatusPart will also work for PowerPMAC (this also depends on https://github.com/dls-controls/pmac/pull/91)


thomascobb (approved):
This looks fine, please can you add release notes showing the minimum version of pmac required, then we can merge

lyodaom:
Done

PR #349, opened 2021-09-27
Remove rank padding on secondary and attribute datasets

GDYendell:
This makes it possible to plot with these datasets on the x axis. Requires [ADCore#474](https://github.com/areaDetector/ADCore/pull/474)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,8 +3,19 @@ Change Log
 All notable changes to this project will be documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`_ after 2-1.
 
+`5.4`_ - 2022-03-30
+-------------------
+
+Changed:
+
+- Removed rank padding on secondary and attribute datasets. Before this change,
+  AreaDetector drivers using the HDF5 file writer would pad these data with an
+  additional 1x1 dimension on top of the scan shape. This makes it simpler for
+  plotting and visualisation. This change requires a minimum ADCore version of
+  3.12 for the HDF5 file writing plugins.
+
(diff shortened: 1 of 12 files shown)


thomascobb (commented):
How do we transition between old and new ADCore? Can we detect it via version number, or should we add a parameter to the HDF writer?

GDYendell:
We could use $(P)$(R)ADCoreVersion_RBV, either on DRV or the HDF5 plugin.

If Mark would like it to be a config option then we can set the PV. Maybe that is cleaner anyway?

BenBradnick (commented):
Looks good. If the ADCore PR is merged as-is then I guess all AreaDetector drivers would need updating to the new ADCore version on a beamline before deploying.

A comment in the changelog on this is needed and it would be useful for it to error if it finds an old ADCore version (in a similar way to PmacTrajectoryPart for the motion program version).

GDYendell:
Does the version check need to go in code, or can we something like the `required_version` config that `DetectorDriverPart` uses?

BenBradnick:
I feel like there's a verb missing there - do you mean move check_driver_version to ADCore/util.py and re-use the same logic in the HDFWriterPart, or just use it as the actual check?

I think it will need to be a separate check in the HDFWriterPart. You could be sending the arrays from the IOC running the AreaDetector driver to another IOC with the HDF5 plugin using the PVAccess plugin, and each of these may not be running against the same version of ADCore.

GDYendell:
I have rebased on the latest. I think this is ready once ADCore 3.12 is released, so we can do the version check. The last commit will need amending when that happens.

PR #360, opened 2021-12-08
Fix/pmacchildpart profile refactor

### LINKED ISSUE #358: Large number of points results in CA timeouts on child parts during configure
As the number of points in a scan increases, so does the configure time taken by a few of the parts. At some point a limit is reached where the pmacChildPart takes so long to configure that it blocks other threads for longer than the timeouts used on Channel Access requests using cothread.catools. Then when those threads start running again they throw timeout errors.

The limit on points varies on the type of generator (e.g. Static or Line with moving axes) and the performance of the host that Malcolm is running on. On P45-control you will get errors at ~2 million static points, but p45-ws001 doesn't error up to 4 million points.

On p45-ws001 the following generator:
```
{
  typeid: 'scanpointgenerator:generator/CompoundGenerator:1.0',
  generators: [
    {
      typeid: 'scanpointgenerator:generator/StaticPointGenerator:1.0',
      size: 8000000
    }
  ],
  excluders: [],
  mutators: [],
  duration: 0.2,
  continuous: true
}
```
Results in the following timeouts (usually multiple PVs across multiple parts):
```
  File "/dls_sw/work/R3.14.12.7/ioc/BL45P/pymalcolm/malcolm/modules/ADCore/parts/statspluginpart.py", line 98, in on_configure
    context.wait_all_futures(fs)
  File "/dls_sw/work/R3.14.12.7/ioc/BL45P/pymalcolm/malcolm/core/context.py", line 318, in wait_all_futures
    self._service_futures(filtered_futures, until)
  File "/dls_sw/work/R3.14.12.7/ioc/BL45P/pymalcolm/malcolm/core/context.py", line 426, in _service_futures
    raise response.message
  File "/dls_sw/work/R3.14.12.7/ioc/BL45P/pymalcolm/malcolm/core/controller.py", line 145, in _handle_request
    responses += handler(request)
  File "/dls_sw/work/R3.14.12.7/ioc/BL45P/pymalcolm/malcolm/core/controller.py", line 204, in _handle_put
    result = put_function(value)
  File "/dls_sw/work/R3.14.12.7/ioc/BL45P/pymalcolm/malcolm/modules/ca/parts/cabooleanpart.py", line 41, in caput
    self.caa.caput(int(value))
  File "/dls_sw/work/R3.14.12.7/ioc/BL45P/pymalcolm/malcolm/modules/ca/util.py", line 271, in caput
    throw=self.throw,
  File "/scratch/cvl62853/pipenv/pymalcolm-PZqGxTx3/lib/python3.7/site-packages/cothread/catools.py", line 761, in caget
    return caget_one(pvs, **kargs)
  File "/scratch/cvl62853/pipenv/pymalcolm-PZqGxTx3/lib/python3.7/site-packages/cothread/catools.py", line 130, in throw_wrapper
    return function(pv, *args, **kargs)
  File "/scratch/cvl62853/pipenv/pymalcolm-PZqGxTx3/lib/python3.7/site-packages/cothread/catools.py", line 648, in caget_one
    return ca_timeout(done, timeout, pv)
  File "/scratch/cvl62853/pipenv/pymalcolm-PZqGxTx3/lib/python3.7/site-packages/cothread/catools.py", line 157, in ca_timeout
    py23.raise_from(ca_nothing(name, cadef.ECA_TIMEOUT), timeout)
  File "<string>", line 4, in raise_from
cothread.catools.ca_nothing: BL45P-EA-DIFF-01:STAT:ComputeStatistics_RBV: User specified timeout on IO operation expired
```

BenBradnick:
Refactored the profile generation method, now iterating through batches and splitting the method into two based on the type of motion trigger to reduce the number of checks.

Also added a yield after each batch so that we don't block other threads for the entire calculation, causing other parts to fail during configure from CA timeouts.

A couple of rough benchmarks on p45-ws001 for start of row triggering:

| Num points  | Configure time before | Configure time after |
| ---------------- | ----------------------------- | -------------------------- |
| 1,000,000 | 1.5s | 1.1s |
|  15,000,000 | 37s  | 22s |

This fixes #358

### DIFF
--- a/malcolm/modules/pmac/parts/pmacchildpart.py
+++ b/malcolm/modules/pmac/parts/pmacchildpart.py
@@ -2,6 +2,7 @@ import re
 from enum import Enum
 from typing import Dict, List, Optional, Union
 
+import cothread
 import numpy as np
 from annotypes import add_call_types
 from scanpointgenerator import CompoundGenerator, Point
@@ -432,9 +433,9 @@ class PmacChildPart(builtin.parts.ChildPart):
         # "how far through the pmac trajectory" rather than a generator
         # scan step
         if scanned > 0:
-            completed_steps = self.completed_steps_lookup[scanned - 1]
             # Only report progress if we are triggering for every point
             if self.output_triggers == scanning.infos.MotionTrigger.EVERY_POINT:
+                completed_steps = self.completed_steps_lookup[scanned - 1]
                 self.registrar.report(scanning.infos.RunProgressInfo(completed_steps))
             # Keep PROFILE_POINTS trajectory points in front
             if (
@@ -690,6 +691,7 @@ class PmacChildPart(builtin.parts.ChildPart):
         or not used at all. Indexing into points[] is expensive so this is
         intentional.
         """
+        profile_point_added = False
         if self.time_since_last_pvt > 0 and not points_are_joined:
             # assume we can skip if we are at the end of a row and we
             # just skipped the most recent point (i.e. time_since_last_pvt > 0)
@@ -712,6 +714,7 @@ class PmacChildPart(builtin.parts.ChildPart):
                 {name: point.positions[name] for name in self.axis_mapping},
             )
             self.time_since_last_pvt = point.duration / 2.0
+            profile_point_added = True
 
         # only add the lower bound if we did not skip this point OR if we are
         # at the end of a row where we always require a final point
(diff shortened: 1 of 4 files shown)


malcolm/modules/pmac/parts/pmacchildpart.py
```diff
@@ -759,6 +760,141 @@ def get_some_points(self, start_index):
...
         return points, joined, velocities
 
+    def check_profile_length_exceeds_profile_points(self) -> bool:
+        # Check if we have exceeded the points number and need to write
+        # Strictly less than so we always add one more point to the time
+        # array so we can always stretch points in a subsequent add with
+        # the values already in the profiles
+        return len(self.profile["timeArray"]) > PROFILE_POINTS
+
+    def create_generator_profile_every_point(self, start_index: int) -> bool:
```
thomascobb, line 770: This looks quite similar to `create_generator_profile_sparse()`. How much code do they share? Could the be re-combined to a single function or does that make it unreadable? If not, could some common code be extracted into other functions?

PR #361, opened 2021-12-15
Feature/hdf5 plugin progress

### LINKED ISSUE #323: HDFWriterPart does not track frames written to disk
Currently scans with AreaDetector HDF5 plugins that fail (e.g. due to permission errors or invalid configuration parameters) will succeed as long as all of the other parts run as expected. I believe instead the scan should fail with an appropriate error.

In my case the value of frames cached per chunk was too large for the frame chunking dimensions, so the writer stopped before the first frame was written.

I think there should be two checks:

- Monitoring for any early errors. The capture_RBV PV will go low if an error occurs, and with defer being True will happen during the first frame, and I think just checking for the first frame to be written should capture most runtime errors.
- Checking at the end that the correct number of frames (NumCaptured_RBV) matches what we expect.

BenBradnick:
This fixes #323.

Some changes to how the HDFWriterPart works:

- Still waits indefinitely for the first frame to arrive (by unique ID number)
- But then uses number of frames captured rather than uniqueID to update completed steps and check for the plugin stalling
- Improved the error message when the writing does stall
- Fixes the first array Future to handle multiple run batches/resuming from pause. Added kwargs to when_value_matches to allow this (so the next expected unique ID can be passed to the callable for the created When instance)

I have tested this on P45 with 2D and 3D scans (software outer scannable) and with pausing.

### DIFF
--- a/malcolm/modules/ADCore/parts/hdfwriterpart.py
+++ b/malcolm/modules/ADCore/parts/hdfwriterpart.py
@@ -558,12 +558,11 @@ class HDFWriterPart(builtin.parts.ChildPart):
                 # This is ok, means we aren't done yet, so flush
                 self._flush_if_still_writing(child)
                 # Check it hasn't been too long since the last frame was written
-                if self.last_capture_update:
-                    if self._has_file_writing_stalled():
-                        timeout_message = self._get_file_writing_stalled_error_message(
-                            child
-                        )
-                        raise TimeoutError(timeout_message)
+                if self._has_file_writing_stalled():
+                    timeout_message = self._get_file_writing_stalled_error_message(
+                        child
+                    )
+                    raise TimeoutError(timeout_message)
             else:
                 break
 
@@ -576,10 +575,12 @@ class HDFWriterPart(builtin.parts.ChildPart):
             # Flush the hdf frames to disk
             child.flushNow()
 
-    def _has_file_writing_stalled(self):
-        return time.time() > self.last_capture_update + self.frame_timeout
+    def _has_file_writing_stalled(self) -> bool:
+        if self.last_capture_update:
+            return time.time() > self.last_capture_update + self.frame_timeout
+        return False
 
-    def _get_file_writing_stalled_error_message(self, child):
+    def _get_file_writing_stalled_error_message(self, child) -> str:
         unique_id = child.uniqueId.value
         num_captured = child.numCapturedReadback.value - self.num_captured_offset
         frames_expected = self.done_when_captured - self.num_captured_offset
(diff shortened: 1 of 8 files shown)


malcolm/modules/ADCore/parts/hdfwriterpart.py
```diff
@@ -41,8 +41,8 @@
 APartRunsOnWindows = APartRunsOnWindows
 
 
-def greater_than_zero(v: int) -> bool:
-    return v > 0
+def greater_than_value(v: int, compare_value: int = 0) -> bool:
+    return v > compare_value
```
thomascobb, line 45: If we write as
```suggestion
def greater_than_value(compare_value: int = 0) -> Callable[[int], bool]:
    def greater_than(v: int) -> bool:
        return v > compare_value
    greater_than.__name__ = f"greater_than_{compare_value)}"
    return greater_than
```
then can we avoid passing kwargs to `When`?

malcolm/modules/ADCore/parts/hdfwriterpart.py
```diff
@@ -516,38 +522,62 @@ def _check_xml_is_valid(self, child):
...
         # will skip to a uniqueID that has not been produced yet
-        self.uniqueid_offset = completed_steps - self.done_when_reaches
-        self.done_when_reaches += steps_to_do
+        self.uniqueid_offset = completed_steps - self.last_expected_unique_id
+        self.last_expected_unique_id += steps_to_do
+        # Update the first array future
+        child = context.block_view(self.mri)
+        self.first_array_future = child.when_value_matches_async(
+            "arrayCounterReadback",
+            greater_than_value,
```
thomascobb, line 537: We could zero the arrayCounter here if that makes the logic easier?

thomascobb (commented):
This looks reasonable. So we no longer compare uniqueID number, but look at the arrayCounter to see if we're finished?

BenBradnick:
> This looks reasonable. So we no longer compare uniqueID number, but look at the arrayCounter to see if we're finished?

Kind of - the arrayCounter is used just to wait for the first array to arrive in the plugin.

Then instead of using uniqueID it looks at the numCaptured_RBV which is the actual number of frames written to the HDF5 file. This is because if the file writing falls over for any reason, it will now be captured (as the file will be missing frames) rather than still succeeding due to having receiving the last frame with the correct uniqueID.

thomascobb:
In which case this looks fine. I marginally prefer my `greater_than` implementation to the `**kwargs`, but there's not much in it

malcolm/modules/ADCore/parts/hdfwriterpart.py
BenBradnick: I think this is a great idea, would allow me to remove some other attributes too

malcolm/modules/ADCore/parts/hdfwriterpart.py
BenBradnick: I wasn't a huge fan of my implementation either, and do like this one. However resetting the array counter on seek to zero avoids needing to change the method from the original which is even nicer!

PR #372, opened 2022-05-18
Add check of soft limit at start ramp and tail-off

BenBradnick:
This adds checks for soft limits in the pmacChildPart when calculating the move to the start position and the tail off at the end of a scan, and raises an Exception if they go out of bounds.

It is expected that GDA will do a basic check of the scan grid/line against the soft limits before submitting the scan to Malcolm.

The remaining edge cases are checking turnaround points and when fly scanning, which stretches out the points.

The other issues to potentially resolve are:

- Handling fly scans which stretch out the points
- Turnaround points
- Should the trajectory be aborted if running when attempting to add the tail off, or whether we can just get away with not adding the tail off point

### DIFF
--- a/malcolm/modules/pmac/infos.py
+++ b/malcolm/modules/pmac/infos.py
@@ -2,8 +2,6 @@ from malcolm.core import Info
 
 from .velocityprofile import VelocityProfile
 
-import math
-
 
 class PmacVariablesInfo(Info):
     """List variable values for variables required in kinematics
@@ -55,6 +53,8 @@ class MotorInfo(Info):
         units: str,
         user_high_limit: float,
         user_low_limit: float,
+        dial_high_limit: float,
+        dial_low_limit: float,
     ) -> None:
         self.cs_axis = cs_axis
         self.cs_port = cs_port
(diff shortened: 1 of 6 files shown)


malcolm/modules/pmac/infos.py
```diff
@@ -109,3 +115,16 @@ def in_cts(self, position: float) -> int:
         """Return the position (in EGUs) translated to counts"""
         cts = int(round((position - self.offset) / self.resolution))
         return cts
+
+    def check_position_within_soft_limits(self, position: float) -> bool:
+        """Check a position (in EGUs) against the soft limits and return True/False"""
+        # Soft limits of 0.0, 0.0 on a motor record means we should ignore them
+        if math.isclose(
```
coretl, line 122: Do we need to do isclose? I though 0.0 was always 0.0 in floating point land? Also, does 0.0, 0.0 disable limits in drive limits or user limits?

BenBradnick: 0.0 does seem to be 0.0 in floating point land - I am guessing it's safe because there's no operation on the value before comparison.

I now use dial limits to determine whether the soft limits are enabled or not.

PR #376, opened 2022-06-15
Feature/ad tucsen

tomtrafford:
Added blocks for using ADTucsen detectors. This currently requires a sleep during on_configure, due to the detector library being ready to acquire before the detector is.

### DIFF
--- a/malcolm/modules/ADTucsen/parts/tucsendriverpart.py
+++ b/malcolm/modules/ADTucsen/parts/tucsendriverpart.py
@@ -1,0 +1,41 @@
+from typing import Any
+from annotypes import Anno, Any, add_call_types
+import cothread
+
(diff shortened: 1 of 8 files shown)


malcolm/modules/ADTucsen/parts/tucsendriverpart.py
```diff
@@ -0,0 +1,43 @@
...
+        super(TucsenDriverPart, self).on_configure(
+            context,
+            completed_steps,
+            steps_to_do,
+            part_info,
+            generator,
+            fileDir,
+            **kwargs,
+        )
+        cothread.Sleep(1.5)
```
coretl, line 43: Are there any PVs we can check instead of doing a sleep?

PR #377, opened 2022-06-21
Ensure Seq Prescale value is not set during a scan

tomtrafford:
If the Prescale parameter on either Sequencer table in the PandA design is set, it will cause the trigger timings to be incorrect. This change ensures that it is 0 during a scan, and logs a warning that the design has been changed by Malcolm.

This change also logs a message when Malcolm changes the value of Repeats, as this must be set to '1'.

### DIFF
--- a/malcolm/modules/ADPandABlocks/doublebuffer.py
+++ b/malcolm/modules/ADPandABlocks/doublebuffer.py
@@ -242,14 +242,14 @@ class DoubleBuffer:
 
                 if table.repeats.value != 1:
                     logging.warning(
-                        "Repeats on Seq table {} should be set to '1', Malcolm has"
-                        "overwritten value for this scan".format(table.label.value)
+                        f"Repeats on Seq table {table.label.value} should be set to "
+                        "'1', Malcolm has overwritten value for this scan"
                     )
                     table.repeats.put_value(1)
                 if table.prescale.value != 0:
                     logging.warning(
-                        "Prescale on Seq table {} should be set to '0', Malcolm has"
-                        "overwritten value for this scan".format(table.label.value)
+                        f"Prescale on Seq table {table.label.value} should be set to "
+                        "'0', Malcolm has overwritten value for this scan"
                     )
                     table.prescale.put_value(0)
             self._finished = False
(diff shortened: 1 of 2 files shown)


malcolm/modules/ADPandABlocks/doublebuffer.py
```diff
@@ -238,7 +239,19 @@ def configure(self, rows_generator: Iterator[SequencerRows]) -> None:
...
             for table in self._table_map.values():
-                table.repeats.put_value(1)
+
+                if table.repeats.value != 1:
+                    logging.warning(
+                        "Repeats on Seq table {} should be set to '1', Malcolm has"
+                        "overwritten value for this scan".format(table.label.value)
+                    )
+                    table.repeats.put_value(1)
+                if table.prescale.value != 0:
```
coretl, line 249: 0 or 1 are fine

tomtrafford: If it is set to 1 it will still scale the timings by the Prescale Units (s, ms, or us). I think the values being inputted to the table are in Clock Ticks, so even when prescale is 1 us the timings do not work.

coretl: Good point, I'd forgotten about units, 0 or 1 clock ticks are fine, but the units complicates checking. I agree that it makes sense to require a value of 0.

malcolm/modules/ADPandABlocks/doublebuffer.py
```diff
@@ -238,7 +239,19 @@ def configure(self, rows_generator: Iterator[SequencerRows]) -> None:
             # The template sets repeats to 1 however older designs may not have
             # been updated.
             for table in self._table_map.values():
-                table.repeats.put_value(1)
+
+                if table.repeats.value != 1:
+                    logging.warning(
+                        "Repeats on Seq table {} should be set to '1', Malcolm has"
```
BenBradnick, line 245: Instead of `.format(table.label.value)` you can use f-strings (https://realpython.com/python-f-strings/):
```suggestion
                       f"Repeats on Seq table {table.label.value} should be set to '1', Malcolm has"
```

tomtrafford: You do know that you no longer need to review these, right :) ? But thanks, I think it looks tidier with f-strings.

PR #380, opened 2022-08-17
Odin VDS fixes

GDYendell:


GDYendell:
I am not sure what branch I based this off of originally but it is gone now. this PR should only add the final commit when it is ready. @EdWarrick have these changes made it into master?

PR #385, opened 2023-01-23
Removed unused ADPilatus param from ADAndor3

adhowell:


malcolm/modules/ADAndor3/blocks/ADAndor3_runnable_block.yaml
```diff
@@ -29,10 +29,6 @@
     mri: $(mri_prefix)
     config_dir: $(config_dir)
 
-- ADPilatus.blocks.ADPilatus_driver_block:
```
tomtrafford, line 32: I think rather than removing the ADPilatus_driver_block we needed to change this to the ADAndor3_driver_block

tomtrafford (approved):

PR #390, opened 2024-06-13
Push PandA layout down to hardware on put

coretl:
This stores the layout table on PandA as `*METADATA.LAYOUT`. Format is:

`{block.name: {"x": block.x, "y": block.y} for block in blocks if block.visible}`

E.g.
`{"COUNTER1": {"x": 1.2, "y": 2.3}, "TTLIN1": {"x": 0.0, "y": 5.6}}`

### DIFF
--- a/malcolm/modules/pandablocks/controllers/pandamanagercontroller.py
+++ b/malcolm/modules/pandablocks/controllers/pandamanagercontroller.py
@@ -274,7 +274,7 @@ class PandAManagerController(builtin.controllers.ManagerController):
             elif field_name == "LAYOUT" and v:
                 # Set the layout table first so that related fields are set after
                 # possible deletion
-                self._json_layout = json.loads(v)
+                self._json_layout = json.loads("".join(v))
                 x, y, visible = [], [], []
                 for name in self.layout.value.name:
                     visible.append(name in self._json_layout)
--- a/malcolm/modules/pandablocks/controllers/pandamanagercontroller.py
+++ b/malcolm/modules/pandablocks/controllers/pandamanagercontroller.py
@@ -1,4 +1,5 @@
 import json
+import re
 import time
 from typing import Any, Dict, Sequence, Set, Tuple
 
@@ -325,6 +326,6 @@ class PandAManagerController(builtin.controllers.ManagerController):
             else:
                 self._json_layout.pop(name, "")
         if self._json_layout != old_json_layout:
-            self._client.set_table(
-                "*METADATA", "LAYOUT", [json.dumps(self._json_layout)]
-            )
+            # Custom encoding so the lines aren't too long and there aren't too many of them
+            lines = re.split(r'(?<=,) (?!"y")', json.dumps(self._json_layout))
+            self._client.set_table("*METADATA", "LAYOUT", lines)
--- a/malcolm/modules/pandablocks/controllers/pandamanagercontroller.py
+++ b/malcolm/modules/pandablocks/controllers/pandamanagercontroller.py
@@ -1,4 +1,3 @@
-import copy
 import json
 import time
 from typing import Any, Dict, Sequence, Set, Tuple
--- a/malcolm/modules/pandablocks/controllers/pandamanagercontroller.py
+++ b/malcolm/modules/pandablocks/controllers/pandamanagercontroller.py
@@ -319,6 +319,7 @@ class PandAManagerController(builtin.controllers.ManagerController):
 
     def set_layout(self, value):
         super().set_layout(value)
+
         json_layout = {}
         for name, _, x, y, visible in self.layout.value.rows():
             if visible:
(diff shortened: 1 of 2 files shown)


coretl:
@tomtrafford this works as a test, but I haven't tried it on a real PandA. It needs building into web-control and putting onto a PandA, then running an updated Malcolm against it, and checking that layout changes in the PandA web GUI and Malcolm web GUI propogate to and from each other...

tomtrafford:
There seems to be a problem with the layout datatype when trying this on an actual PandA. I get "

> ERR Cannot write to this field

". Which gets raised by the PandAserver here: https://github.com/PandABlocks/PandABlocks-server/blob/ee4c35271426aa949b2b7001edd367687f7524f6/server/metadata.c#L161C10-L161C28 
It looks like it is being sent as a string whereas it needs to be sent as a Multiline (or a metadata_table?). But I don't really know how the server interprets inputs in different ways.

I can replicate this and what should happen when talking directly through the tcp-server.
```
< *METADATA.LAYOUT={"COUNTER1": {"x": 23.0, "y": -199.5}}
> ERR Cannot write to this field

< *METADATA.LAYOUT<
< {"COUNTER1": {"x": 23.0, "y": -199.5}}
< 
> OK
```

But I don't know how to edit the Malcolm interaction to match the 2nd example.

coretl:
I think this commit will make it write it as a multiline table...

malcolm/modules/pandablocks/controllers/pandamanagercontroller.py
```diff
@@ -267,6 +271,20 @@ def _handle_change(self, k, v, bus_changes, block_changes, bit_out_changes):
         if block_name == "*METADATA":
             if field_name.startswith("LABEL_"):
                 field_name, block_name = field_name.split("_", 1)
+            elif field_name == "LAYOUT" and v:
+                # Set the layout table first so that related fields are set after
+                # possible deletion
+                self._json_layout = json.loads(v)
```
tomtrafford, line 277:
```suggestion
                self._json_layout = json.loads("".join(v))
```
v is a list but this expects a string.

tomtrafford:
I've had a quick test of this on a PandA and it works. I can see changes made on the PandA design on Malcolm and visa-versa. I've encountered a couple of possible issues:

- When Malcolm starts it clears the layout.
- There are some elements on the Malcolm design which don't exist on the PandA. For example, the POS and HDF blocks - They disappear instantly when I added them to the Malcolm layout. When I attempted to load the double_seq_pcomp template it caused a broken pipe error between malcolm and the PandA and Malcolm had to be restarted to recover.

coretl:
Ok, I'll try and take a look at this later today

coretl:
@tomtrafford does this fix it?

tomtrafford:
That has fixed the problem -I can now put Malcolm only blocks on the layout. However I'm seeing some more problems at the moment: When the layout gets too big Malcolm gets an error: `malcolm.core.errors.TimeoutError: Queue().get() timed out
` From a little testing it looks like this occurs when *METADATA.LAYOUT is around ~1000 characters long (around 18 blocks).

I've also found that I cannot write to the exports table with these changes. The list of available sources is blank.

coretl:
> That has fixed the problem -I can now put Malcolm only blocks on the layout. However I'm seeing some more problems at the moment: When the layout gets too big Malcolm gets an error: `malcolm.core.errors.TimeoutError: Queue().get() timed out ` From a little testing it looks like this occurs when *METADATA.LAYOUT is around ~1000 characters long (around 18 blocks).

@Araneidae is this a limit in the TCP server?

Araneidae:
Actually, I'm wondering if the `MAX` here:

    size_t new_length = MAX(MIN_LENGTH, 2 * writer->max_length);

should simply be `MIN`.  Given the rest of the code, I suspect so...

coretl:
Ok, I've split each block metadata to be on a different line, so that shouldn't hit the 1024 character limit. @tomtrafford does this fix it? Is the export list still not working?

tomtrafford:
This seems to have fixed the 1024 character limit problem now. The export list still does not work - the list of sources available to be exported is empty. If I try to load a design which has signals in the exports table I get an error: `Array(['SEQ1.table', 'SEQ2.table', 'SRGATE2.forceSet', 'SRGATE2.forceRst']) is not a valid value in Array([]) for element 0`

Araneidae:
I've checked the behaviour of the Panda Server when an overlong command line is received, and what it does is:
* Closes the requesting socket
* Reports `raised error: Line overrun` to the log

Does this match what you were seeing?

Perhaps the server could be fixed to treat line overruns differently, one option would be to discard characters until a newline is seen and then report the "Line overrun" to the connected client.

tomtrafford:
Yes that matches what I am seeing. I can see the following message in the log:
```
Jun 14 12:29:13 (none) daemon.err PandA server: Client config 172.23.245.108:58302 raised error: Line overrun
Jun 14 12:29:13 (none) daemon.info PandA server: Client config 172.23.245.108:58302 closed
```

